PR #7, opened 2024-11-29
Control plane testing overhaul

roshanrags:
Complete overhaul of control plane testing infrastructure with the goal of
- easier to write tests
- clearer tests
- more complete tests (exhaustive checks)

Summary of changes:
- Common test runner so tests are more clear and concise without noise
- Outcomes are compared as vectors so it checks count, order and all outcome fields exhaustively
- `SystemContext` to wrap the start time global so tests have separate instances and run without interfering with each other
- `UpdateEnclaveImageOutcome` to track `update_enclave_image` calls explicitly
- Randomness replaced with deterministic derivations
- Some comments calling out (to me) surprising behaviour

Tests themselves and coverage can be improved as well but that's for another day.

### DIFF
--- a/operator/control-plane/src/market.rs
+++ b/operator/control-plane/src/market.rs
@@ -19,13 +19,13 @@ use tracing::{error, info, info_span, Instrument};
 
 // Trait to encapsulate behaviour that should be simulated in tests
 trait SystemContext {
-    fn now_timestamp() -> Duration;
+    fn now_timestamp(&self) -> Duration;
 }
 
 struct RealSystemContext {}
 
 impl SystemContext for RealSystemContext {
-    fn now_timestamp() -> Duration {
+    fn now_timestamp(&self) -> Duration {
         use std::time::SystemTime;
         SystemTime::now()
             .duration_since(SystemTime::UNIX_EPOCH)
--- a/operator/control-plane/src/market.rs
+++ b/operator/control-plane/src/market.rs
@@ -1478,7 +1478,7 @@ mod tests {
         let job_id = format!("{:064x}", 1);
 
         let logs = vec![
-            (0, Action::Open, ("{\"region\":\"ap-south-1\",\"url\":\"https://example.com/enclave.eif\",\"instance\":\"c6a.xlarge\",\"memory\":4096,\"vcpu\":2}".to_string(),31000000000000u64,31000u64,market::now_timestamp().as_secs()).abi_encode_sequence()),
+            (0, Action::Open, ("{\"region\":\"ap-south-1\",\"url\":\"https://example.com/enclave.eif\",\"instance\":\"c6a.xlarge\",\"memory\":4096,\"vcpu\":2}".to_string(),31000000000000u64,31000u64,0).abi_encode_sequence()),
             (301, Action::Close, [].into()),
         ];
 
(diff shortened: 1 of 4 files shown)


operator/control-plane/src/market.rs
```diff
@@ -1400,1166 +1433,882 @@ mod tests {
...
                 chain: "123".into(),
             },
-            &["ap-south-1".into()],
-            300,
-            &test::get_rates(),
-            &test::get_gb_rates(),
-            &Vec::from([
+            allowed_regions: vec!["ap-south-1".to_owned()],
+            address_whitelist: vec![
                 "0x0000000000000000000000000f5f91ba30a00bd43bd19466f020b3e5fc7a49ec".to_string(),
```
vg-27, line 2134: we can parameterize this address during test setup. And provide the same to test::getLogs(). It was't clear what value should this address match to, since getLogs uses some default address.

operator/control-plane/src/market.rs
```diff
@@ -2573,10 +2322,8 @@ mod tests {
...
     #[tokio::test]
     async fn test_whitelist_blacklist_check_whitelisted() {
-        let _ = market::START.set(Instant::now());
-
         let log = test::get_log(Action::Open,
-            Bytes::from(("{\"region\":\"ap-south-1\",\"url\":\"https://example.com/enclave.eif\",\"instance\":\"c6a.xlarge\",\"memory\":4096,\"vcpu\":2}".to_string(),31000000000000u64,31000u64,market::now_timestamp().as_secs()).abi_encode_sequence()),
+            Bytes::from(("{\"region\":\"ap-south-1\",\"url\":\"https://example.com/enclave.eif\",\"instance\":\"c6a.xlarge\",\"memory\":4096,\"vcpu\":2}".to_string(),31000000000000u64,31000u64,0).abi_encode_sequence()),
             B256::ZERO);
         let address_whitelist = vec![
             "0x0000000000000000000000000f5f91ba30a00bd43bd19466f020b3e5fc7a49ec".to_string(),
```
vg-27, line 2329: here also address can be parameterized.

operator/control-plane/src/market.rs
```diff
@@ -2753,178 +2490,128 @@ mod tests {
...
-        let _ = market::START.set(Instant::now());
+        let start_time = Instant::now();
+        let job_id = format!("{:064x}", 1);
 
-        let job_num = U256::from(1).into();
-        let job_logs: Vec<(u64, Log)> = vec![
-            (0, Action::Open, ("{\"region\":\"ap-south-1\",\"url\":\"https://example.com/enclave.eif\",\"instance\":\"c6a.xlarge\",\"memory\":4096,\"vcpu\":2}".to_string(),31000000000000u64,31000u64,market::now_timestamp().as_secs()).abi_encode_sequence()),
+        let logs = vec![
+            (0, Action::Open, ("{\"region\":\"ap-south-1\",\"url\":\"https://example.com/enclave.eif\",\"instance\":\"c6a.xlarge\",\"memory\":4096,\"vcpu\":2}".to_string(),31000000000000u64,31000u64,0).abi_encode_sequence()),
             // instance type has also been updated in the metadata. should fail this job.
```
vg-27, line 2593: remove this comment, there is not change in instance type. Also, in this test, instance should be running instead of failing since there isn't any update, please check.

operator/control-plane/src/market.rs
roshanrags: get_log is actually quite messy, it uses the hardcoded address for everything from emitting contract to sender to depositors, should probably rework that

vg-27 (dismissed):

vg-27:
LGTM

PR #10, opened 2024-12-05
Read docker image tar files from local

Naman-Monga:


sdks/docker-enclave/Dockerfile
roshanrags, line 1: What is this doing?

sdks/docker-enclave/docker-compose-start.sh
roshanrags, line 1: Loading should probably be moved to setup.sh

sdks/docker-enclave/default.nix
```diff
@@ -10,6 +10,8 @@
   vet,
   kernels,
   compose ? ./. + "/docker-compose.yml",
+  dockerImageTar ? ./. + "/docker-image.tar",
```
roshanrags, line 13: should be an array to pass in multiple tars

sdks/docker-enclave/supervisord.conf
```diff
@@ -43,15 +43,19 @@ stderr_logfile_maxbytes=0
 # docker
 [program:docker]
 command=/bin/dockerd --iptables=false
+priority=10
```
roshanrags, line 46: the version of supervisord we use has a `depend_on` setting for dependencies

ref: https://github.com/ochinchina/supervisord?tab=readme-ov-file#supervised-program-settings

sdks/docker-enclave/Dockerfile
Naman-Monga: put this as an example. but cool, will remove it, not needed.

sdks/docker-enclave/docker-compose-start.sh
Naman-Monga: Since setup.sh doesn't have the docker daemon running. I thought it'd be better to load image and run the docker-image in a single script, instead of running dockerd twice.  your thoughts on this?

roshanrags: the supervisor program is supposed to be restartable on failure, can't load every time it restarts

Naman-Monga: done.

sdks/docker-enclave/default.nix
Naman-Monga: done

sdks/docker-enclave/default.nix
```diff
@@ -10,6 +10,7 @@
   vet,
   kernels,
   compose ? ./. + "/docker-compose.yml",
+  dockerImagesDir ? ./. + "/docker-images",
```
roshanrags, line 13: I think make it list of paths, easier to handle once we have a cli tool wrapper

sdks/docker-enclave/Dockerfile
roshanrags, line 1: remove

sdks/docker-enclave/supervisord.conf
```diff
@@ -40,19 +40,23 @@ stderr_logfile_maxbytes=0
 
 # your custom programs go here
 
-# docker
-[program:docker]
+# Docker daemon
+[program:dockerd]
 command=/bin/dockerd --iptables=false
 autorestart=true
+startretries=3
```
roshanrags, line 47: what is this for?

sdks/docker-enclave/default.nix
Naman-Monga: done.

sdks/docker-enclave/supervisord.conf
Naman-Monga: was initially using this snippet for the build script which checked if docker-daemon has started and loaded images. will remove it for this.

Naman-Monga: but I believe keeping it for the docker compose will be a good to have, just incase some sorta resource allocation fails. thoughts?
```
# Docker Compose
[program:compose]
command=docker compose -f /app/docker-compose.yml up --remove-orphans
autorestart=true
startretries=3
stdout_logfile=/dev/stdout
stdout_logfile_maxbytes=0
stderr_logfile=/dev/stdout
stderr_logfile_maxbytes=0
depends_on=dockerd
```

roshanrags: I think it's fine, let it keep trying, not like stopping has any point when the enclave is still occupied

roshanrags:
Superseded by #15

PR #18, opened 2024-12-16
Enforce serverless output limit

vg-27:
Executor will return error code 5 upon computing the serverless output that is larger than size limit.

### DIFF
--- a/serverless/executor/src/job_handler.rs
+++ b/serverless/executor/src/job_handler.rs
@@ -232,7 +232,7 @@ async fn execute_job(
             error_code: 5,
             total_time: execution_timer_start.elapsed().as_millis().into(),
             ..Default::default()
-        })
+        });
     }
 
     Some(JobOutput {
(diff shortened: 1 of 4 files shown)


serverless/executor/src/job_handler.rs
```diff
@@ -224,6 +226,15 @@ async fn execute_job(
         return None;
     };
 
+    if response.len() > MAX_OUTPUT_BYTES_LENGTH {
+        return Some(JobOutput{
+            output: Bytes::new(),
+            error_code: 5,
+            total_time: execution_timer_start.elapsed().as_millis().into(),
+            ..Default::default()
+        })
```
Ayush-Yadav, line 235: nit: add ; at the end

Ayush-Yadav (dismissed):

roshanrags:
`MAX_OUTPUT_BYTES_LENGTH` has to be consistent across the network right?

serverless/executor/src/job_handler.rs
vg-27: Done

roshanrags (approved):

PR #20, opened 2024-12-17
[Serverless Executor] Inject encrypted gas wallet private key

vg-27:
Previously, gas wallet private key was set using http insecure calls, where private key was exposed.

After this change, enclave owner can inject private key securely. Also, gas private key will be accepted only if signed by the configured owner to prevent misconfiguration of gas private key.


vg-27:
Closing this PR. This PR https://github.com/marlinprotocol/oyster-monorepo/pull/26 resolves the secure key injection.

PR #30, opened 2024-12-24
Remove pcr verification and just print decoded fields

roshanrags:


sdks/rs/src/main.rs
```diff
@@ -10,44 +10,31 @@ struct Cli {
...
     let decoded = verify(
         attestation_doc,
         AttestationExpectations {
             age: Some((cli.max_age, now)),
-            pcrs: Some(pcrs),
             ..Default::default()
         },
     )?;
     println!("verification successful: {:?}", decoded);
+    println!("pcr0: {}", hex::encode(decoded.pcrs[0]));
```
vg-27, line 33: we are already printing the decoded variable, can remove that if not required.

vg-27 (approved):
Rest LG.

PR #32, opened 2024-12-25
[Serverless] Replace HTTP server with HTTP over vsock

vg-27:
Http over vsock secures the communication between serverless operator and the enclave. Configurations are made securely over this connection. Connections can only originate from the nitro enclave host, thus preventing any other external access.

### DIFF
--- a/serverless/http-on-vsock-client/src/gw_vsock_client.rs
+++ b/serverless/http-on-vsock-client/src/gw_vsock_client.rs
@@ -10,13 +10,17 @@ struct Cli {
     #[clap(short, long, value_parser)]
     url: String,
 
-    // owner address
+    /// owner address
     #[clap(short, long, value_parser)]
     owner_address: String,
 
-    // gas key
+    /// gas key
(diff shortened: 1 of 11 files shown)


serverless/http-on-vsock-client/src/gw_vsock_client.rs
```diff
@@ -0,0 +1,92 @@
...
+use http_on_vsock_client::vsock_client::vsock_connector;
+use hyper::{body::Body, Request, Uri};
+use serde_json::json;
+
+#[derive(Parser)]
+#[clap(author, version, about, long_about = None)]
+struct Cli {
+    /// url to query
+    #[clap(short, long, value_parser)]
+    url: String,
```
arnav-deep, line 11: hostname or domain might be a clearer name?

vg-27: argument looks like a url only `vsock://<cid>:<port>/`

arnav-deep (approved):

PR #33, opened 2024-12-25
[Transaction Manager] Nonce Too High Error and more example clarity.

arnav-deep:
### Nonce Too High Error
- Ensure any nonce set for a transaction is utilised, even if sending that transaction fails using a dummy transaction.
- Add Gas Price and Qty Estimation in dummy transaction.

### Example Clarity
- ContractInstance of `alloy-rs` can be used to create function call struct and converted into bytes to get the transaction data for that fn call.

### Contract Execution Error while Gas Estimation
- ContractExecution error during gas estimation does not stop the fn call and keeps on trying. Stop retries and send a dummy transaction to utilise nonce.

### DIFF
--- a/serverless/transaction-manager/src/errors.rs
+++ b/serverless/transaction-manager/src/errors.rs
@@ -12,6 +12,8 @@ pub enum TxnManagerSendError {
     NonceTooHigh(String),
     #[error("Out of gas. Error: {0}")]
     OutOfGas(String),
+    #[error("Not enough gas. Error: {0}")]
+    NotEnoughGas(String),
     #[error("Gas too high. Error: {0}")]
     GasTooHigh(String),
     #[error("Gas price low. Error: {0}")]
--- a/serverless/transaction-manager/src/errors.rs
+++ b/serverless/transaction-manager/src/errors.rs
@@ -13,7 +13,7 @@ pub enum TxnManagerSendError {
     #[error("Out of gas. Error: {0}")]
     OutOfGas(String),
     #[error("Not enough gas. Error: {0}")]
-    NotEnoughGas(String),
+    InsufficientBalance(String),
     #[error("Gas too high. Error: {0}")]
     GasTooHigh(String),
     #[error("Gas price low. Error: {0}")]
--- a/serverless/transaction-manager/src/errors.rs
+++ b/serverless/transaction-manager/src/errors.rs
@@ -12,7 +12,7 @@ pub enum TxnManagerSendError {
     NonceTooHigh(String),
     #[error("Out of gas. Error: {0}")]
     OutOfGas(String),
-    #[error("Not enough gas. Error: {0}")]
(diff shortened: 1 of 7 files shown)


vg-27 (commented):
Changes looks good. Need testing results before merging.

serverless/transaction-manager/src/utils.rs
```diff
@@ -18,6 +18,7 @@ pub(crate) fn parse_send_error(error: String) -> TxnManagerSendError {
     if error_lowercase.contains("out of gas")
         || (error_lowercase.contains("transaction requires at least")
             && error_lowercase.contains("gas but got"))
+        || error_lowercase.contains("gas too low")
```
vg-27, line 21: is this  is the error that you got from sepolia rpc?

serverless/transaction-manager/src/utils.rs
```diff
@@ -18,6 +18,7 @@ pub(crate) fn parse_send_error(error: String) -> TxnManagerSendError {
     if error_lowercase.contains("out of gas")
         || (error_lowercase.contains("transaction requires at least")
```
vg-27, line 19: btw, I noticed this, this error occurs when, when gas account has less ETHs than required, am I right? If yes, then we can remove it from here and create another error type.

vg-27 (commented):
LG

serverless/transaction-manager/src/utils.rs
arnav-deep: Yes

serverless/transaction-manager/src/utils.rs
arnav-deep: Updated to a new type

serverless/transaction-manager/src/errors.rs
```diff
@@ -12,6 +12,8 @@ pub enum TxnManagerSendError {
     NonceTooHigh(String),
     #[error("Out of gas. Error: {0}")]
     OutOfGas(String),
+    #[error("Not enough gas. Error: {0}")]
+    NotEnoughGas(String),
```
vg-27, line 16: can we name it `InsufficientBalance` ?

serverless/transaction-manager/src/errors.rs
```diff
@@ -13,7 +13,7 @@ pub enum TxnManagerSendError {
     #[error("Out of gas. Error: {0}")]
     OutOfGas(String),
     #[error("Not enough gas. Error: {0}")]
```
vg-27, line 15: update the string too

vg-27 (approved):
LG

serverless/transaction-manager/src/errors.rs
arnav-deep: Done

vg-27 (approved):
LG

PR #35, opened 2024-12-25
Cli bug fixes

Naman-Monga:


cli/oyster-cvm/src/main.rs
```diff
@@ -34,12 +34,16 @@ enum Commands {
         docker_compose: String,
 
         /// List of Docker image .tar file paths
-        #[arg(short = 'i', long, required = true)]
-        docker_images: Vec<String>,
+        #[arg(short = 'i', long, required = false)]
```
roshanrags, line 37: can just set `default_value=""` equivalent, avoids Option everywhere

roshanrags (approved):

cli/oyster-cvm/src/main.rs
```diff
@@ -36,12 +36,20 @@ enum Commands {
...
 
         /// Output folder name
         #[arg(short, long, default_value = "result")]
         output: String,
+
+        /// Git commit reference for oyster-monorepo
+        #[arg(
+            short = 'r',
+            long,
+            default_value = "3e6dbc844b42281462e65d7742d9436d4205fcfd"
```
vg-27, line 50: why is this exact commit used as default? need comment to instruct if this needs to be changed in future. When does a commit qualify for a default value here.

cli/oyster-cvm/src/commands/build.rs
```diff
@@ -1,46 +1,97 @@
...
+        let line = line_result.context("Failed to read line from stdout")?;
+        info!(target: "nix", "{}", line);
+    }
+
+    info!("Build finished");
+
+    Ok(())
+}
+
+fn check_trusted_user() -> Result<()> {
```
vg-27, line 65: this function is repeated from doctor, can be included in common util

cli/oyster-cvm/src/commands/build.rs
```diff
@@ -1,46 +1,97 @@
...
 
 pub fn build_oyster_image(
     platform: Platform,
     docker_compose: &str,
     docker_images: &[String],
     output: &str,
+    commit_ref: &str,
 ) -> Result<()> {
+    info!("Checking if the current user is in the nix trusted-users list");
+    check_trusted_user()?;
```
vg-27, line 15: why are we using this check here? what happens with nix command if this is not checked?

vg-27 (approved):

cli/oyster-cvm/src/main.rs
Naman-Monga: any commit with the current nix configs will be relevant. Will have to be changed incase we make any changes to nix configs (add any param / change the project structure).
Will add a comment in the next pr.

cli/oyster-cvm/src/commands/build.rs
Naman-Monga: Actually this was was added as the error wasn't readable if trusted user was failing, but with nix user would instantly know. 
But yeah, to make things more organized, we can add a dependency checker, which scans for all things necessary in the env for that specific command.

cli/oyster-cvm/src/main.rs
Naman-Monga: done

cli/oyster-cvm/src/commands/build.rs
Naman-Monga: removed from build, this will only run with doctor.

PR #39, opened 2024-12-27
Catch new 'gas price low' error description from the RPC

Ayush-Yadav:
Catch the new error description from RPC for gas price low: `max fee per gas less than block base fee`


vg-27 (approved):
LG. What will be the test steps with arb sepolia RPC?

Ayush-Yadav:
> LG. What will be the test steps with arb sepolia RPC?

Testing done on sepolia with the RPC error parser categorizing this error description under `GasPriceLow` and handling the transaction sending accordingly with setting increased gas price for resending.

PR #54, opened 2025-01-06
Add prebuilt binaries

roshanrags:


vg-27 (approved):
LG. Does this section of README will contain prebuilt binaries of future releases/tags too or the latest one only?

PR #58, opened 2025-01-07
[SERVERLESS GATEWAY] Refactor gateway cycle state

arnav-deep:
### Gateway Cycle State
- Add a sleep before all RPC provider calls inside a loop.
- Use block rate in next block calculation for optmisation
- Loop around RPC call for address information and retry in case error.

### Gateway VScock Client
- Fix: Errors are not json, print body directly.

### DIFF
--- a/serverless/gateway/src/chain_util.rs
+++ b/serverless/gateway/src/chain_util.rs
@@ -15,7 +15,7 @@ use std::future::Future;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
 use std::time::{SystemTime, UNIX_EPOCH};
-use tokio::time;
+use tokio::time::{sleep, Duration};
 
 use crate::constant::{
     MAX_RETRY_ON_PROVIDER_ERROR, MAX_TX_RECEIPT_RETRIES, WAIT_BEFORE_HTTP_RPC_CALL,
@@ -84,7 +84,7 @@ pub async fn get_block_number_by_timestamp(
                 "Failed to fetch block number. Error: {:#?}",
                 get_block_number_result.err()
             );
-            time::sleep(time::Duration::from_millis(WAIT_BEFORE_HTTP_RPC_CALL)).await;
+            sleep(Duration::from_millis(WAIT_BEFORE_HTTP_RPC_CALL)).await;
             continue;
         }
 
@@ -113,10 +113,12 @@ pub async fn get_block_number_by_timestamp(
                 block_number,
                 block.err()
             );
+            sleep(Duration::from_millis(WAIT_BEFORE_HTTP_RPC_CALL)).await;
             continue;
         }
         let block = block.unwrap();
         if block.is_none() {
+            sleep(Duration::from_millis(WAIT_BEFORE_HTTP_RPC_CALL)).await;
             continue;
         }
         let block = block.unwrap();
@@ -148,13 +150,14 @@ pub async fn get_block_number_by_timestamp(
                             block_number = earliest_block_number_after_target_ts - 1;
                             earliest_block_number_after_target_ts -= 1;
                         }
+                        sleep(Duration::from_millis(WAIT_BEFORE_HTTP_RPC_CALL)).await;
                         continue 'less_than_block_number;
                     }
                     Ok(None) => {
                         // The next block does not exist.
                         // Wait for the next block to be created to be sure that
                         // the current block_number is the required block_number
-                        time::sleep(time::Duration::from_millis(WAIT_BEFORE_HTTP_RPC_CALL)).await;
+                        sleep(Duration::from_millis(WAIT_BEFORE_HTTP_RPC_CALL)).await;
                         continue 'next_block_check;
                     }
                     Err(err) => {
@@ -164,6 +167,7 @@ pub async fn get_block_number_by_timestamp(
                         );
                         retry_on_error += 1;
                         if retry_on_error <= MAX_RETRY_ON_PROVIDER_ERROR {
+                            sleep(Duration::from_millis(WAIT_BEFORE_HTTP_RPC_CALL)).await;
                             continue 'next_block_check;
                         }
                         return None;
@@ -439,7 +443,7 @@ pub async fn confirm_event(
                         log.removed = true;
                         break;
                     }
-                    time::sleep(time::Duration::from_millis(WAIT_BEFORE_HTTP_RPC_CALL)).await;
+                    sleep(Duration::from_millis(WAIT_BEFORE_HTTP_RPC_CALL)).await;
                     continue;
                 }
             };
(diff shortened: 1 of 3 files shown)


serverless/gateway/src/common_chain_gateway_state_service.rs
```diff
@@ -319,26 +319,35 @@ pub async fn generate_gateway_epoch_state_for_cycle(
     let gateway_addresses: Vec<Address> = current_cycle_state_epoch.keys().cloned().collect();
 
     for address in gateway_addresses {
-        let gateways_info = com_chain_gateway_contract
-            .gateways(address)
-            .block(BlockId::from(to_block_number))
-            .call()
-            .await;
+        for _ in 0..MAX_RETRY_ON_PROVIDER_ERROR {
```
vg-27, line 322: upon failing to obtain gateway info from contracts, we are just skipping there state update. This will lead to inconsistency of random selection across gateways. Unrelated to objective of this PR/issue. Really need to look into it.

serverless/http-on-vsock-client/src/gw_vsock_client.rs
```diff
@@ -100,8 +100,13 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     println!("{:?}", resp);
     let body_bytes = hyper::body::to_bytes(resp.into_body()).await?;
     let body_str = String::from_utf8(body_bytes.to_vec())?;
-    let json: serde_json::Value = serde_json::from_str(&body_str)?;
-    println!("{:?}", json);
+
+    if resp.status().is_success() {
```
vg-27, line 104: is this tested?

vg-27 (changes requested):

serverless/gateway/src/chain_util.rs
```diff
@@ -98,7 +98,7 @@ pub async fn get_block_number_by_timestamp(
     }
 
     // A conservative estimate of the block rate per second before it is actually calculated below.
```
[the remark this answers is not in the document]
arnav-deep: Done

serverless/gateway/src/chain_util.rs
```diff
@@ -140,21 +147,33 @@ pub async fn get_block_number_by_timestamp(
...
                             return Some(block_number);
                         }
-                        block_number = block_number
-                            + ((target_timestamp - block.header.timestamp) as f64
-                                * block_rate_per_second) as u64;
+
+                        if block.header.timestamp != first_block_timestamp
+                            && block.header.timestamp + 1 < target_timestamp
+                        {
+                            block_rate_per_second = (first_block_number - next_block_number) as f64
```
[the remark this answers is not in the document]
arnav-deep: Yeah. Modified this a bit.

serverless/gateway/src/common_chain_gateway_state_service.rs
arnav-deep: The logic before this did the same. There was no retry then, so a retry was added now. Other than network error, there's no reason for this read method to give an error here.

serverless/http-on-vsock-client/src/gw_vsock_client.rs
arnav-deep: Yes. Modified a bit

serverless/gateway/src/chain_util.rs
```diff
@@ -148,11 +147,17 @@ pub async fn get_block_number_by_timestamp(
                             return Some(block_number);
                         }
 
-                        if block.header.timestamp != first_block_timestamp
+                        if block.header.timestamp != first_block_number
```
vg-27, line 150: isn't this wrong? comparing time with block number?

serverless/gateway/src/chain_util.rs
```diff
@@ -148,11 +147,17 @@ pub async fn get_block_number_by_timestamp(
...
                             && block.header.timestamp + 1 < target_timestamp
                         {
-                            block_rate_per_second = (first_block_number - next_block_number) as f64
-                                / (first_block_timestamp - block.header.timestamp) as f64;
+                            if block.header.timestamp < first_block_timestamp {
```
vg-27, line 153: is this case possible? Aren't we putting lower bound on search space for blocks just like upper limit.

serverless/gateway/src/chain_util.rs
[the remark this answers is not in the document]
vg-27: > Can we increase the range of calculating the block rate? Like this function will be called periodically by epoch state generation, so it can get block of last epoch as input and use it as first block?

This isn't addressed?

serverless/gateway/src/chain_util.rs
arnav-deep: Yes. Have addressed this.

serverless/gateway/src/chain_util.rs
arnav-deep: This is possible. In most cases, the first_block_timestamp is the latest block timestamp. So first_block_timestamp might be greater than the target_timestamp, so we reduce the block number. But here, the first block number is already set as the latest. So depending on what cycle it is, it can be greater, or larger, and the block rate is very dynamic, so there are cases that after the block rate calculation, it puts the `block_number` to a block which is greater than the first block, and then in the next iteration, the block rate is calculated again, and it puts the `block_number` to the less than first block block. Due to the dynamicity of the block rate, the block_number goes to greater and then smaller and then less greater and then less smaller, and eventually to the exact number. It's still fewer iterations.

serverless/gateway/src/chain_util.rs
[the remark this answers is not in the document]
arnav-deep: Currently no. Was checking the performance of this approach. Will add this and compare

serverless/gateway/src/chain_util.rs
```diff
@@ -159,10 +159,15 @@ pub async fn get_block_number_by_timestamp(
                                     / (block.header.timestamp - first_block_timestamp) as f64;
                             }
                             info!("Block rate per second: {}", block_rate_per_second);
-                            block_number = block_number
+                            let new_block_number = block_number
                                 + ((target_timestamp - block.header.timestamp) as f64
                                     * block_rate_per_second)
                                     as u64;
+                            if new_block_number == block_number {
```
vg-27, line 166: this case can occur only if
- target_timestamp == block.header.timestamp (which is ruled out because of outer if condition)
- block_rate_per_second == 0  ==> next_block_number == first_block_number ==> same timestamp (which is also ruled out by the outer if condition)

So, is this case really occurring?

serverless/gateway/src/chain_util.rs
```diff
@@ -159,10 +159,15 @@ pub async fn get_block_number_by_timestamp(
                                     / (block.header.timestamp - first_block_timestamp) as f64;
```
vg-27, line 159: there is a bug at L158, you should have subtracted first_block_number.

vg-27 (changes requested):

vg-27 (approved):

vg-27 (approved):
LG

roshanrags (approved):

PR #72, opened 2025-01-16
Simpler scallop + user data in decoded attestation

roshanrags:
Removed stuff like Pcrs from the scallop interface, the implementation can just pick how and what it persists. E.g. implementations might want to use `user_data`, or non-enclave attestations for that matter.

### DIFF
--- a/sdks/rs/src/scallop.rs
+++ b/sdks/rs/src/scallop.rs
@@ -180,9 +180,9 @@ pub type UserData = Box<[u8]>;
 
 pub trait ScallopAuthStore {
     fn contains(&self, key: &Key) -> bool;
-    fn get(&self, key: &Key) -> Option<&Pcrs>;
-    fn set(&mut self, key: Key, pcrs: Pcrs);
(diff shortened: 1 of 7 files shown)


sdks/rs/src/scallop.rs
```diff
@@ -175,48 +175,42 @@ enum ReadMode {
...
 pub type Key = [u8; 32];
-pub type Pcrs = [[u8; 48]; 3];
+#[derive(PartialEq)]
+pub enum ContainsResponse {
+    // the key was found and is approved
+    Approved,
+    // the key was not found
+    NotFound,
+    // the key is rejected
+    Rejected,
```
vg-27, line 185: What would be example scenario where a key is marked rejected? Is this something like blacklisting it manually?

vg-27 (approved):
LG

sdks/rs/src/scallop.rs
roshanrags: You could have explicit blacklisting, but you could also have keys dynamically cached as blacklisted after verification or even later. It basically just exits the protocol early before getting into attestation verification and such.

PR #74, opened 2025-01-20
Integrate secret store with serverless executor

Ayush-Yadav:
* Update secret store API endpoints
* Update serverless executor API endpoints
* Integrate secret store endpoints with executor APIs
* Update job execution logic to take into account the secret to be injected


serverless/secret-store/src/server.rs
```diff
@@ -131,12 +147,18 @@ async fn inject_mutable_config(
                 err
             ));
         }
+
+        let mut ws_rpc_url = app_state.web_socket_url.write().unwrap();
```
vg-27, line 151: Don't Repeat Yourself (DRY), make a separate function to do this. Also, can if moved outside of if else, since it is common part?

serverless/secret-store/src/server.rs
```diff
@@ -106,6 +116,12 @@ async fn inject_mutable_config(
...
 
         *app_state.http_rpc_txn_manager.lock().unwrap() = Some(http_rpc_txn_manager);
+        let mut ws_rpc_url = app_state.web_socket_url.write().unwrap();
+        // strip existing api key from the ws url by removing keys after last '/'
+        let pos = ws_rpc_url.rfind('/').unwrap();
+        ws_rpc_url.truncate(pos + 1);
+        ws_rpc_url.push_str(mutable_config.ws_api_key.as_str());
+
         *mutable_params_injected_guard = true;
         drop(mutable_params_injected_guard);
```
vg-27, line 126: we should the drop the locks in reverse order of acquiring them. Is it required to drop this lock here explicitly? Note that you still has hold of RW lock on web_socket_url, which should be dropped first.

serverless/secret-store/src/server.rs
```diff
@@ -131,12 +147,18 @@ async fn inject_mutable_config(
...
+        let pos = ws_rpc_url.rfind('/').unwrap();
+        ws_rpc_url.truncate(pos + 1);
+        ws_rpc_url.push_str(mutable_config.ws_api_key.as_str());
     }
 
     HttpResponse::Ok().body("Mutable params configured!\n")
 }
 
-#[get("/store-details")]
+#[post("/store-details")]
```
vg-27, line 161: why is it marked post?

serverless/executor/src/workerd.rs
```diff
@@ -144,6 +146,9 @@ const oysterWorker :Workerd.Worker = (
   modules = [
     (name = \"main\", esModule = embed \"{tx_hash}-{slug}.js\")
   ],
+  bindings = [
+    (name = \"secret\", data = embed \"{secret_store_path}/{secret_id}.bin\")
```
vg-27, line 150: let's check a case where two jobs are running in parallel using the same secret. Need to check if secret file can be used in parallel. One job should not wait for another job to consume the secret.

serverless/executor/src/main.rs
```diff
@@ -107,7 +109,7 @@ async fn main() -> Result<()> {
         .route("/", get(index))
         .route("/immutable-config", post(inject_immutable_config))
         .route("/mutable-config", post(inject_mutable_config))
-        .route("/executor-details", get(get_executor_details))
+        .route("/tee-details", get(get_tee_details))
```
vg-27, line 112: update the http vsock client file too with new endpoint.

serverless/executor/src/node_handler.rs
```diff
@@ -57,6 +49,37 @@ pub async fn inject_immutable_config(
...
+    let request_json = json!({
+        "owner_address_hex": immutable_config.owner_address_hex,
+    });
+
+    let secret_store_response = Retry::spawn(
+        ExponentialBackoff::from_millis(5).map(jitter).take(3),
+        || async { call_secret_store_endpoint(&secret_store_endpoint, request_json.clone()).await },
+    )
+    .await;
+    let Ok(_) = secret_store_response else {
```
vg-27, line 63: why is the handling of status code from the secret store is skipped ? what if it is an failed status code?

serverless/executor/src/node_handler.rs
```diff
@@ -171,13 +216,24 @@ pub async fn inject_mutable_config(
     let pos = ws_rpc_url.rfind('/').unwrap();
     ws_rpc_url.truncate(pos + 1);
     ws_rpc_url.push_str(mutable_config.ws_api_key.as_str());
-    *mutable_params_injected_guard = true;
 
-    (StatusCode::OK, "Mutable params configured!\n").into_response()
+    if secret_store_response.0 == reqwest::StatusCode::OK {
```
vg-27, line 220: can you handle it just after the http call to secret store? Any specific reason to check it at the end?

serverless/secret-store/src/server.rs
```diff
@@ -159,9 +181,9 @@ async fn get_secret_store_details(app_state: Data<AppState>) -> impl Responder {
     }))
 }
 
-#[get("/signed-registration-message")]
+#[post("/signed-registration-message")]
```
vg-27, line 184: update endpoint name also as similar to function name

serverless/executor/src/node_handler.rs
```diff
@@ -312,3 +445,55 @@ pub async fn export_signed_registration_message(app_state: State<AppState>) -> R
 
     (StatusCode::OK, Json(response_body)).into_response()
 }
+
+async fn call_secret_store_endpoint(
+    endpoint: &str,
```
vg-27, line 450: it's very unconventional that endpoint include both port and endpoint after that. You can just add separate arg for port, and prepare url in this function.

serverless/executor/src/node_handler.rs
```diff
@@ -226,6 +314,49 @@ pub async fn export_signed_registration_message(app_state: State<AppState>) -> R
             .into_response();
     }
 
+    let secret_store_endpoint =
+        app_state.secret_store_config_port.to_string() + "/signed-registration-message";
+    let request_json = json!({});
+
+    let secret_store_response = Retry::spawn(
+        ExponentialBackoff::from_millis(5).map(jitter).take(3),
```
vg-27, line 322: adding these retries in the helper function would make more sense. Also, normally retries should be added for those errors which can be resolved after retry not all errors. Try to avoid unnecessary retries.

serverless/executor/src/node_handler.rs
```diff
@@ -312,3 +445,55 @@ pub async fn export_signed_registration_message(app_state: State<AppState>) -> R
 
     (StatusCode::OK, Json(response_body)).into_response()
 }
+
+async fn call_secret_store_endpoint(
```
vg-27, line 449: move this function to util?

vg-27 (changes requested):

serverless/secret-store/src/server.rs
```diff
@@ -52,6 +47,12 @@ async fn inject_immutable_config(
         return HttpResponse::BadRequest().body("Owner address must be 20 bytes long!\n");
     }
 
+    let mut immutable_params_injected_guard = app_state.immutable_params_injected.lock().unwrap();
```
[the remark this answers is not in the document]
Ayush-Yadav: There was no specific reason, have made the APIs a bit more structured in the sense that if the parameters are invalid then we don't proceed further into taking locks and all

serverless/secret-store/src/server.rs
Ayush-Yadav: Yes dropping the lock is necessary because we have to make an await call to TxnManager afterwards. Will drop the web_socket lock explicitly as well. Thanks for pointing out.

serverless/executor/src/workerd.rs
Ayush-Yadav: Sure will check this case

serverless/executor/src/node_handler.rs
Ayush-Yadav: Even if setting mutable params in the store fails, we should still proceed to try setting it for the executor. The last check is just to mark if mutable params have been injected in both store and executor or not which will allow us to export registration message for the enclave.

serverless/secret-store/src/server.rs
Ayush-Yadav: Yeah can move this part before if else statement

serverless/secret-store/src/server.rs
vg-27: actually, we should run the txn_manager when we start the listner for events, not in this function.

serverless/secret-store/src/server.rs
vg-27: these methods have their own purpose. Read upon GET, POST, PUT, DELETE methods. Here since this is only returning the details (like read method), GET should be there.

serverless/executor/src/node_handler.rs
vg-27: so, if owner is already configured in secret store then check the owner and if it is not intended then fail the API. Can't have inconsistency. But this got me into thinking why we need owner address for secret store in this case where executor is getting integrated with the secret store. Can remove that API at all, right?

serverless/executor/src/node_handler.rs
vg-27: 1. If we return http error at L209, i.e. just after secret store api call failure, what problem will it cause?
2. If secret store api fails, then user has to set mutable config again, right?

serverless/executor-enclave/default.nix
```diff
@@ -22,6 +23,7 @@
   attestationServer = "${attestation-server}/bin/oyster-attestation-server";
   keygenSecp256k1 = "${keygen}/bin/keygen-secp256k1";
   executor' = "${executor}/bin/oyster-serverless-executor";
+  secret-store' = "${secret-store}/bin/oyster-secret-store";
```
vg-27, line 26: use camelCase

serverless/executor/src/event_handler.rs
```diff
@@ -18,7 +18,9 @@ use tokio_retry::strategy::{jitter, ExponentialBackoff};
 use tokio_retry::Retry;
 use tokio_stream::Stream;
 
+use crate::constant::*;
```
vg-27, line 21: using explicit imports instead of * is better import practice.

serverless/executor/src/main.rs
```diff
@@ -14,11 +14,9 @@ use k256::ecdsa::SigningKey;
 use tokio::fs;
 
 use serverless::cgroups::Cgroups;
-use serverless::node_handler::{
-    export_signed_registration_message, get_tee_details, index, inject_immutable_config,
-    inject_mutable_config,
-};
-use serverless::utils::{load_abi_from_file, AppState, ConfigManager};
+use serverless::model::{AppState, ConfigManager};
+use serverless::node_handler::*;
```
vg-27, line 18: keep the explicit imports

vg-27 (changes requested):

serverless/executor/src/node_handler.rs
Ayush-Yadav: We need the owner address in secret store to capture the secret store registered event so can't remove that.

serverless/executor/src/event_handler.rs
Ayush-Yadav: All the constants in the constant.rs file are needed in the event_handler so have added this.

serverless/executor/src/main.rs
Ayush-Yadav: All the endpoints in node_handler are required here to start the servers so have added this.

serverless/executor/src/node_handler.rs
Ayush-Yadav: 1. There is no problem, but we can return the error after setting the executor mutable params also if possible since mutable params for both are independent and can be needed to set again for them separately. 
2. Yes right. If secret store api fails then we'll return the error in the end and won't set mutable_params_injected to true if not already.

serverless/executor/src/node_handler.rs
vg-27: > so, if owner is already configured in secret store then check the owner and if it is not intended then fail the API. Can't have inconsistency.

what about this?

serverless/executor/src/node_handler.rs
vg-27: why do you want to set the mutable param in executor even if secret store api fails? Given, user has to call the api again.

Also, you are creating inconsistency in the executor state. That mutable_params_injected_guard is false but the ws_api, nonce_to_send and http_rpc_client has been set.

I want to keep this invariability.

vg-27 (commented):
Waiting for tests.

serverless/secret-store/src/user_utility/sign_secret.rs
```diff
@@ -57,25 +104,212 @@ fn main() -> Result<()> {
...
+    if U256::from(secret_data_bytes.len()) > decoded_log.sizeLimit {
+        return Err(anyhow!("Secret data length exceeds limit"));
+    }
+
+    if decoded_log
+        .selectedEnclaves
+        .iter()
+        .any(|add| !config.stores.contains_key(add))
+    {
+        return Err(anyhow!("Enclave info doesn't exist for a selected enclave"));
```
vg-27, line 187: you can print all missing enclave address whose info is not present.

serverless/secret-store/src/user_utility/sign_secret.rs
```diff
@@ -57,25 +104,212 @@ fn main() -> Result<()> {
...
+    }
+
+    let mut injection_tasks = vec![];
+
+    for addr in decoded_log.selectedEnclaves {
+        let store_info = config.stores.get(&addr).unwrap().clone();
+        let secret_data_bytes_clone = secret_data_bytes.clone();
+        let user_private_key_clone = user_private_key.clone();
+        injection_tasks.push(tokio::spawn(async move {
```
vg-27, line 196: try something like this https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html#method.join_all

serverless/secret-store/src/user_utility/sign_secret.rs
```diff
@@ -57,25 +104,212 @@ fn main() -> Result<()> {
...
+async fn inject_secret(
+    store_info: EnclaveInfo,
+    secret_id: U256,
+    secret_data_bytes: Vec<u8>,
+    user_private_key: SigningKey,
+) {
+    let enclave_public_key = hex::decode(store_info.public_key);
+    let Ok(enclave_public_key) = enclave_public_key else {
+        eprintln!(
+            "Failed to hex decode the enclave public key: {:?}\n",
```
vg-27, line 223: add the enclave address for which it is failing. for other errors too.

serverless/secret-store/src/user_utility/sign_secret.rs
```diff
@@ -57,25 +104,212 @@ fn main() -> Result<()> {
...
+        );
+        return;
+    };
+
     println!(
-        "Secret ID: {}\nEncrypted secret: {}\nSignature: {}",
-        cli.secret_id,
-        hex::encode(encrypted_secret_data_bytes),
-        hex::encode(signature)
+        "Secret injected successfully with acknowledgement: {:?}",
```
vg-27, line 311: print enclave address too.

serverless/secret-store/src/user_utility/sign_secret.rs
```diff
@@ -1,51 +1,98 @@
...
-    #[clap(long, value_parser)]
-    enclave_public_key: String,
-
     // User's private key to sign the data
     #[clap(long, value_parser)]
     user_private_hex: String,
+
+    // Store enclaves info configuration
+    #[clap(long, value_parser)]
+    config_file: String,
```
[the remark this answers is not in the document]
Ayush-Yadav: Default enclaves?

serverless/executor/src/tests.rs
```diff
@@ -1340,25 +1672,98 @@ pub mod serverless_executor_test {
...
     }
 
+    async fn mock_post_endpoint(port: u16, endpoint: &str) -> Arc<Mutex<(StatusCode, String)>> {
+        let shared_state = Arc::new(Mutex::new((StatusCode::OK, String::new())));
+
+        let state_clone = shared_state.clone();
+        let app = Router::new().route(
+            endpoint,
+            post(move || {
+                let state = state_clone.clone();
```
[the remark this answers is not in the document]
Ayush-Yadav: Yeah it's not needed. Will remove.

serverless/secret-store/src/user_utility/sign_secret.rs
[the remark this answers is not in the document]
vg-27: i meant default value of the file name in cli arg config in clap.

serverless/secret-store/src/user_utility/sign_secret.rs
```diff
@@ -52,45 +62,148 @@ impl ConfigManager {
...
-struct SecretStoredAcknowledgement {
-    secret_id: U256,
-    sign_timestamp: u64,
-    signature: String,
-}
-
 #[derive(Parser)]
 #[command(author, version, about, long_about = None)]
 struct Cli {
     // Http rpc url
```
vg-27, line 68: you can use 3 `///` to show message in help option

serverless/secret-store/src/user_utility/sign_secret.rs
```diff
@@ -179,137 +285,31 @@ async fn main() -> Result<()> {
...
+                addr,
                 store_info,
                 decoded_log.secretId,
                 secret_data_bytes_clone,
                 user_private_key_clone,
             )
             .await
-        }));
+        });
     }
```
vg-27, line 310: What will be the user flow, if some of the addresses are missing from config? Because secret would be submitted partially.

Ayush-Yadav: So either the user can update the config with the info of missing addresses and try again, or if they fail to do so under the acknowledgement timeout then stores will send the timeout transaction and the secret will be reverted.

vg-27 (commented):
LG

vg-27 (commented):
LG, it will need UT coverage

serverless/secret-store/src/main.rs
```diff
@@ -121,14 +122,18 @@ async fn main() -> Result<()> {
...
-    })
-    .bind(("0.0.0.0", args.external_port))
-    .context(format!("could not bind to port {}", args.external_port))?
-    .run();
+    let external_server =
+        HttpServer::new(move || {
+            App::new()
+                .app_data(app_data.clone())
+                .app_data(JsonConfig::default().limit(
+                    2 * config.global_max_store_size_bytes + INJECT_SECRET_JSON_PAYLOAD_BUFFER,
```
vg-27, line 130: add some comment about this config. Also rename the `global_max_store_size_bytes`. this is the max size of secret, right?

serverless/secret-store/src/scheduler.rs
```diff
@@ -96,48 +95,39 @@ pub async fn store_alive_monitor(app_state: Data<AppState>) {
         {
             eprintln!("Failed to send store alive transaction: {:?}", err);
         };
+
+        // Call the garbage cleaner
+        garbage_cleaner(app_state.clone()).await;
```
vg-27, line 100: please add comments on top of function that garbage cleaner is called here. Because function name doesn't imply this responsibility.

vg-27 (dismissed):
LG

serverless/executor/src/event_handler.rs
```diff
@@ -83,6 +94,15 @@ pub async fn events_listener(app_state: State<AppState>, starting_block: U64) {
                     event.block_number.unwrap_or(starting_block).as_u64(),
                     Ordering::SeqCst,
                 );
+
+                // Spawn task for monitoring pending transactions for block confirmation and retrying when necessary
+                let app_state_clone = app_state.clone();
+                let pending_txns_clone = pending_txns.clone();
+                let tx_clone = tx.clone();
+                tokio::spawn(async move {
```
vg-27, line 102: what if you spawn this thread just before the outer loop of web socket connection? Just after send_execution_output thread is spawned.

Requirement is that  multiple threads of resent_pending_transaction should not be running. Check all cases that it is getting terminated after deregistration.

serverless/executor/src/event_handler.rs
```diff
@@ -160,29 +180,14 @@ pub async fn events_listener(app_state: State<AppState>, starting_block: U64) {
...
-            send_execution_output(app_state_clone, rx, pending_txns).await;
-        });
 
         handle_event_logs(
             jobs_created_stream,
             jobs_responded_stream,
             executor_deregistered_stream,
             app_state.clone(),
-            tx,
+            tx_clone,
```
vg-27, line 190: can use tx without clone?

serverless/secret-store/src/events.rs
```diff
@@ -85,6 +86,12 @@ pub async fn events_listener(app_state: Data<AppState>, starting_block: u64) {
                     .unwrap();
                 txn_manager.run().await;
 
+                // Spawn task to submit periodic proofs for secrets stored and remove expired secrets
+                let app_state_clone = app_state.clone();
+                tokio::spawn(async move {
```
vg-27, line 91: confirm that multiple instances of this thread is not running after deregister and again register.

serverless/secret-store/src/scheduler.rs
```diff
@@ -30,6 +30,9 @@ pub async fn remove_expired_secrets_and_mark_store_alive(app_state: Data<AppStat
 
         // If enclave is drained, skip the alive transaction because acknowledgments won't be accepted then
         if app_state.enclave_draining.load(Ordering::SeqCst) {
+            // Call the garbage cleaner to clean all secrets stored inside it
+            garbage_cleaner(app_state.clone(), true).await;
```
vg-27, line 34: you can call this from drain event handler instead of calling it periodically here.

serverless/secret-store/src/user_utility/sign_secret.rs
```diff
@@ -284,21 +293,45 @@ async fn main() -> Result<()> {
...
+            input: TransactionInput::new(
+                getSelectedEnclavesCall {
+                    _secretId: decoded_log.secretId,
+                }
+                .abi_encode()
+                .into(),
+            ),
+            ..Default::default()
+        })
+        .await?;
```
vg-27, line 308: this seems to be complicated way to call view function on a contract. can you simplify it? Maybe define a contract object `sol!` macro.

serverless/secret-store/src/user_utility/sign_secret.rs
```diff
@@ -284,21 +293,45 @@ async fn main() -> Result<()> {
...
+    if decoded.is_empty() {
+        return Err(anyhow!("No selected enclaves found!"));
+    }
+
     let mut inject_set = JoinSet::new();
 
-    for addr in decoded_log.selectedEnclaves {
-        if !config.stores.contains_key(&addr) {
-            eprintln!("Enclave with address {} not present in config data", addr);
+    for addr in decoded {
```
vg-27, line 319: here inject_secret will throw error for those enclaves which have the secret stored already, is that correct?

vg-27 (changes requested):

serverless/executor/src/event_handler.rs
Ayush-Yadav: If we spawn the `resent_pending_transaction` there then it will terminate instantly because the enclave is not registered yet. Because of this line: https://github.com/marlinprotocol/oyster-monorepo/blob/bdc6cbd21a4202510c880f9b20ccd6e3ace40b1a/serverless/executor/src/event_handler.rs#L548

serverless/executor/src/event_handler.rs
Ayush-Yadav: Nope. That way it'll be borrowed in the first loop run and the subsequently web socket connection break loop runs can't borrow the sender. Did confirm from the enclave logs that when the web socket connection breaks, the `handle_event_logs` thread terminates and the loop runs again to establish the connection.

serverless/secret-store/src/user_utility/sign_secret.rs
Ayush-Yadav: Will look into it.

serverless/secret-store/src/user_utility/sign_secret.rs
Ayush-Yadav: Yes the injection API of those enclaves will throw `Secret ID not created yet or undergoing injection!`

serverless/executor/src/event_handler.rs
vg-27: And if that line is removed from the function, i.e. it doesn't check for the registered state.

That would be equivalent to the case of txn-manager, that doesn't depend upon enclave registration, right?

Not a blocker for now.

vg-27 (dismissed):
LG

serverless/secret-store/README.md
```diff
@@ -135,12 +135,22 @@ $ curl -X POST -H "Content-Type: application/json" -d '{"secret_id": {ASSIGNED_S
...
+Secret injected successfully into enclave "{ENCLAVE_ADDRESS}" with acknowledgement: Object {"secret_id": String("{SECRET_ID}"), "sign_timestamp": Number({ENCLAVE_SIGN_TIMESTAMP}), "signature": String("{ENCLAVE_ACKNOWELDGEMENT_SIGNATURE}")}
+...
 ```
+The "SECRET_STORE_ENCLAVES_INFO_JSON" will look something like this: 
+```
+{
+    "stores": {
+        "{ENCLAVE_ADDRESS}": {
+            "public_key": "{ENCLAVE_PUBLIC_KEY}",
+            "store_external_ip": "http://{ENCLAVE_PUBLIC_IP}:{EXTERNAL_PORT}"
```
vg-27, line 147: it should be url

serverless/executor/src/event_handler.rs
```diff
@@ -489,6 +511,10 @@ pub async fn handle_event_logs(
                 }
                 app_state.last_block_seen.store(current_block.as_u64(), Ordering::SeqCst);
 
+                if app_state.enclave_draining.load(Ordering::SeqCst) {
```
vg-27, line 514: we should still mark the job completed, which has been stored in map before draining started. Continuing here will leave open jobs in map.

serverless/executor/src/tests.rs
```diff
@@ -1845,6 +1850,166 @@ pub mod serverless_executor_test {
...
+        app_state.enclave_draining.store(true, Ordering::SeqCst);
+
+        let code_hash = "9c641b535e5586200d0f2fd81f05a39436c0d9dd35530e9fb3ca18352c3ba111";
+        let user_deadline = 5000;
+        let app_state_clone = app_state.clone();
+
+        let code_input_bytes: Bytes = serde_json::to_vec(&json!({})).unwrap().into();
+
+        // Add revive log for the executor
+        let executor_revive_log = vec![Log {
```
vg-27, line 1943: should we drain it first and then revive? That will be correct setup.

serverless/executor/src/tests.rs
```diff
@@ -1845,6 +1850,166 @@ pub mod serverless_executor_test {
...
+            topics: vec![
+                keccak256(EXECUTOR_REVIVED_EVENT).into(),
+                H256::from(app_state.enclave_address),
+            ],
+            removed: Some(false),
+            block_number: Some(1.into()),
+            ..Default::default()
+        }];
+
+        // Add log entry to relay a job but job response event is not sent
```
vg-27, line 1954: update the comment, we are sending the job response.

serverless/executor/src/tests.rs
Ayush-Yadav: Have marked the `enclave_draining` to `true` in the beginning of the UT, Isn't that sufficient? Since we have already tested the draining flow in the previous test case

serverless/executor/src/event_handler.rs
vg-27: Can do.

serverless/executor/src/tests.rs
vg-27: yeah, that's enough

vg-27 (approved):
LG

PR #75, opened 2025-01-23
Fix user code contract comparison by ignoring ascii case 

Ayush-Yadav:


serverless/executor/src/workerd.rs
```diff
@@ -77,7 +77,7 @@ pub async fn create_code_file(
     }?;
 
     // Check if the contract address matches the expected user code contract
-    if contract_address != contract {
+    if contract_address.eq_ignore_ascii_case(contract) {
```
vg-27, line 80: add a UT to cover this case.

vg-27 (approved):
LG. Merge it after testing once with local node.

PR #80, opened 2025-01-24
KMS creator

roshanrags:
Intended to run in an enclave.

It has an endpoint to generate encrypted randomness with two main goals:
1. It is verifiable that the encrypted randomness was generated inside a creator enclave even after the enclave is no longer running.
2. The encrypted randomness can be loaded into KMS root server enclaves.

1 follows from the enclave itself plus the creator signing the encrypted randomness. 2 follows from the condition used by the creator to encrypt the randomness, it can be set to be loadable by enclaves only after verification in the KmsRoot contract.

This will be used to generate a one-time encrypted random seed than can be stored/replicated anywhere and used for key derivations forever.


kms/creator/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "kms-creator"
+version = "0.1.0"
+edition = "2021"
```
vg-27, line 4: add license

vg-27 (commented):
Add README, license and update top README.

vg-27 (commented):
UTs required for this now?

roshanrags:
> UTs required for this now?

IMO skip, don't think there's much UTable here given randomness unless we figure out how to seed both our rng as well the nucypher rng

kms/creator/Cargo.toml
roshanrags: done

kms/creator/LICENSE.txt
```diff
@@ -0,0 +1,678 @@
+Copyright (C) 2024 Marlin Foundation
```
vg-27, line 1: year update?

kms/creator/README.md
```diff
@@ -0,0 +1,66 @@
...
+  -h, --help
+          Print help
+  -V, --version
+          Print version
+```
+
+## Example
+
+```
+$ curl <ip:port>/attestation/raw -vs | xxd
```
vg-27, line 60: update endpoint

vg-27 (dismissed):

vg-27 (commented):
LG, thanks.

vg-27 (approved):

PR #81, opened 2025-01-24
bugfix: vet should handle non utf8 inputs as well

roshanrags:


initialization/vet/src/main.rs
```diff
@@ -107,9 +107,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let response_bytes =
         hyper::body::to_bytes(client.get(cli.url.try_into()?).await?.into_body()).await?;
 
-    let res = String::from_utf8(response_bytes.to_vec())?;
-
-    print!("{res}");
+    tokio::io::stdout().write_all(&response_bytes).await?;
```
vg-27, line 110: so stdout will be always redirected to a file, whenever this program is used, right? Should we take file name as argument to write the response.

vg-27 (approved):
LG. Posted one non-blocker suggestion.

PR #82, opened 2025-01-26
Key derivation utilities

roshanrags:

PR #83, opened 2025-01-26
KMS root server

roshanrags:
Depends on #82


kms/root-server/src/derive_secp256k1_address_ethereum.rs
```diff
@@ -0,0 +1,45 @@
...
+    extract::{Query, State},
+    http::StatusCode,
+};
+use kms_derive_utils::{derive_enclave_seed, derive_path_seed, to_secp256k1_ethereum_address};
+use serde::Deserialize;
+
+use crate::AppState;
+
+#[derive(Deserialize)]
+pub struct Params {
```
vg-27, line 11: same struct seems to be getting used at multiple handles, can use a common struct?

kms/root-server/src/main.rs
```diff
@@ -0,0 +1,249 @@
...
+mod derive_ed25519_public;
+mod derive_secp256k1_address_ethereum;
+mod derive_secp256k1_public;
+mod derive_x25519_public;
+mod scallop;
+mod taco;
+
+#[derive(Parser, Debug)]
+#[command(version, about, long_about = None)]
+struct Args {
```
vg-27, line 33: since there are lot of cli args, instead, can we have a input from json config file, to make it dev friendly.

kms/root-server/src/main.rs
```diff
@@ -0,0 +1,249 @@
...
+#[tokio::main]
+async fn main() -> Result<()> {
+    setup_logging();
+
+    let args = Args::parse();
+
+    // sleep to allow attestation verification
+    // taco decryption will only work after the signer is
+    // verified on chain
+    sleep(Duration::from_secs(args.delay)).await;
```
vg-27, line 97: can replace it with event listener to check if signer is verified on chain?

kms/root-server/src/taco.rs
```diff
@@ -0,0 +1,269 @@
...
+                BASE64_STANDARD.encode(decryption_request.to_bytes()),
+            );
+
+            (provider.to_owned(), shared)
+        })
+        .collect::<HashMap<Address, SessionSharedSecret>>();
+
+    let resp = Client::new().post(porter).json(&porter_req).send().await?;
+    if resp.status() != StatusCode::OK {
+        let bytes = resp.bytes().await?;
```
vg-27, line 193: can  use .text() directly?

kms/root-server/src/taco.rs
```diff
@@ -0,0 +1,269 @@
...
+            if decrypted.ritual_id != ritual_id {
+                return Err(anyhow!("wrong ritual"));
+            }
+
+            Ok(bincode::deserialize::<DecryptionShareSimple>(
+                &decrypted.decryption_share[..],
+            )?)
+        })
+        .filter_map(Result::ok)
+        .collect::<Vec<_>>();
```
vg-27, line 235: should we check length of decrypted secrets against threshold? Since, filter_map is applied.

kms/root-server/src/main.rs
```diff
@@ -0,0 +1,249 @@
...
+            "/derive/ed25519/public",
+            get(derive_ed25519_public::derive_ed25519_public),
+        )
+        .route(
+            "/derive/ed25519/address/solana",
+            get(derive_ed25519_address_solana::derive_ed25519_address_solana),
+        )
+        .route(
+            "/derive/x25519/public",
+            get(derive_x25519_public::derive_x25519_public),
```
vg-27, line 230: what was the reason behind keeping all the handlers in separate files instead of in a single file?

kms/root-server/src/scallop.rs
```diff
@@ -0,0 +1,62 @@
...
+                // TODO: hardcoded, make it a param
+                age: Some((300000, now)),
+                root_public_key: Some(AWS_ROOT_KEY.to_vec()),
+                // do not care about PCRs, will derive different keys for each set
+                ..Default::default()
+            },
+        ) else {
+            return None;
+        };
+
```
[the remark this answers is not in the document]
roshanrags: done

kms/root-server/src/derive_ed25519_address_solana.rs
```diff
@@ -0,0 +1,45 @@
...
+    };
+    let Ok(user_data) = hex::decode(params.user_data) else {
+        return (StatusCode::BAD_REQUEST, String::new());
+    };
+    if user_data.len() > 65535 {
+        return (StatusCode::BAD_REQUEST, String::new());
+    }
+
+    let enclave_key = derive_enclave_seed(state.randomness, &pcr0, &pcr1, &pcr2, &user_data);
+    let path_key = derive_path_seed(enclave_key, params.path.as_bytes());
```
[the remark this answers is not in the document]
roshanrags: no, could be anything

kms/root-server/src/derive_secp256k1_address_ethereum.rs
roshanrags: IMO not worth it rn, might have to add other params in the future

kms/root-server/src/main.rs
roshanrags: in prod, you wouldn't be specifying _any_ args btw, it's just there for our testing

kms/root-server/src/main.rs
roshanrags: contract address is not known beforehand

roshanrags: could make it so one is updated after the other is deployed but that adds admin control where none is really needed

kms/root-server/src/taco.rs
roshanrags: yeah will add a check. have also asked how to handle shares being more than the threshold and if that is fine to just combine all

kms/root-server/src/main.rs
roshanrags: nothing specific, just started that way

kms/root-server/src/taco.rs
roshanrags: current thing is fine

kms/root-server/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+[dependencies]
+alloy = { version = "0.9.2", features = ["full"] }
+anyhow = "1.0.95"
+axum = "0.8.1"
+base64 = "0.22.1"
+bincode = "1.3.3"
+chrono = "0.4.39"
+clap = { version = "4.5.26", features = ["derive"] }
+hex = "0.4.3"
+kms-derive-utils = { version = "0.1.0", path = "../derive-utils" }
```
vg-27, line 16: nix shouldn't be working, because generated cargo.lock won't have any reference to local lib, need to use the github link.

vg-27 (changes requested):
Rest looks good, except the check for nix build.

kms/root-server/Cargo.toml
roshanrags: default.nix has a patch btw, to use the local lib source

vg-27 (approved):
LG.

PR #87, opened 2025-01-28
Derive server enclave

roshanrags:


kms/derive-server-enclave/setup.sh
```diff
@@ -0,0 +1,73 @@
...
+/app/vet --url vsock://3:1300/instance/ip > /app/ip.txt
+cat /app/ip.txt
+
+# query job id for enclave and store
+/app/vet --url vsock://3:1300/oyster/job > /app/job.txt
+cat /app/job.txt
+
+# query init params for enclave and store
+/app/vet --url vsock://3:1300/oyster/init-params > /app/init-params
+/app/vet --url vsock://3:1300/oyster/extra-init-params > /app/extra-init-params
```
vg-27, line 15: We are keeping these since this one will be used as example, right?

vg-27 (approved):
LG

kms/derive-server-enclave/setup.sh
roshanrags: yeah

PR #88, opened 2025-01-28
Creator enclave

roshanrags:
Parameters still correspond to a dry run I'm doing on Polygon. Has to be changed to Ethereum mainnet for final deployment.

### DIFF
--- a/kms/creator-enclave/supervisord.conf
+++ b/kms/creator-enclave/supervisord.conf
@@ -1,0 +1,41 @@
+[supervisord]
+loglevel=debug
+logfile=/dev/stdout
+logfile_maxbytes=0
+
+# attestation server
+[program:attestation-server]
+command=/app/attestation-server --ip-addr 0.0.0.0:1300 --pub-key /app/secp256k1.pub --init-params /app/init-params
+autorestart=true
+stdout_logfile=/dev/stdout
+stdout_logfile_maxbytes=0
+stderr_logfile=/dev/stdout
+stderr_logfile_maxbytes=0
+
+# proxy components inside enclave
+[program:ip-to-vsock-raw-outgoing]
+command=/app/ip-to-vsock-raw-outgoing --vsock-addr 3:1200 --queue-num 0
+autorestart=true
+stdout_logfile=/dev/stdout
+stdout_logfile_maxbytes=0
+stderr_logfile=/dev/stdout
+stderr_logfile_maxbytes=0
+
+[program:vsock-to-ip-raw-incoming]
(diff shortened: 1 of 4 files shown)


kms/creator-enclave/supervisord.conf
```diff
@@ -0,0 +1,41 @@
+[supervisord]
+loglevel=debug
+logfile=/dev/stdout
+logfile_maxbytes=0
+
+# attestation server
+[program:attestation-server]
+command=/app/attestation-server --ip-addr 0.0.0.0:1300 --pub-key /app/secp256k1.pub --init-params /app/init-params
```
vg-27, line 8: there is no /app/init-params , also it is not required for kms creator

kms/creator-enclave/supervisord.conf
```diff
@@ -0,0 +1,41 @@
...
+stdout_logfile=/dev/stdout
+stdout_logfile_maxbytes=0
+stderr_logfile=/dev/stdout
+stderr_logfile_maxbytes=0
+
+# your custom programs go here
+
+# kms creator
+[program:kms-creator]
+command=/app/kms-creator --dkg-public-key 868c3d012a5d524f0939e4ee4d60b738b4c44448ec286a5361e15ffbf2641e2df25363a204a738231e5f1a9621999741 --condition '{"condition":{"chain":137,"conditionType":"contract","contractAddress":"0xB7c6b3aCc6095eB2a8eff8f16B250D442712F093","functionAbi":{"inputs":[{"internalType":"address","name":"","type":"address"}],"name":"isVerified","outputs":[{"internalType":"bool","name":"","type":"bool"}],"stateMutability":"view","type":"function"},"method":"isVerified","parameters":[":userAddress"],"returnValueTest":{"comparator":"==","value":true}},"version":"1.0.0"}'
```
vg-27, line 36: future improvement, better to pass condition as input file.

vg-27 (changes requested):

kms/creator-enclave/supervisord.conf
roshanrags: yeah but the param is a required param, it handles internally even if file doesn't exist

kms/creator-enclave/supervisord.conf
roshanrags: yeah maybe :thinking:

kms/creator-enclave/supervisord.conf
vg-27: can use empty file, or a comment here just to avoid the confusion.

vg-27 (approved):

PR #90, opened 2025-01-28
Upgrade verifier

roshanrags:
RISCZero has a new zkVM version, should upgrade to it for the KMS.

### DIFF
--- a/attestation/verifier-risczero/host/src/main.rs
+++ b/attestation/verifier-risczero/host/src/main.rs
@@ -46,4 +46,10 @@ fn main() {
     let receipt = prove_info.receipt;
 
     println!("{:?}", receipt);
+
+    println!(
+        "Seal: {}",
+        hex::encode(&receipt.inner.groth16().unwrap().seal)
+    );
+    println!("{}", hex::encode(&receipt.journal.bytes));
 }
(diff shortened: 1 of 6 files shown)


attestation/verifier-risczero/host/src/main.rs
```diff
@@ -46,4 +46,10 @@ fn main() {
     let receipt = prove_info.receipt;
 
     println!("{:?}", receipt);
+
+    println!(
+        "Seal: {}",
+        hex::encode(&receipt.inner.groth16().unwrap().seal)
+    );
+    println!("{}", hex::encode(&receipt.journal.bytes));
```
vg-27, line 54: add "journal" before printing it's value

attestation/verifier-risczero/host/src/main.rs
```diff
@@ -46,4 +46,10 @@ fn main() {
     let receipt = prove_info.receipt;
 
     println!("{:?}", receipt);
```
vg-27, line 48: add "Receipt" before printing it's value

roshanrags: done

attestation/verifier-risczero/host/src/main.rs
roshanrags: done

vg-27 (approved):
LG

PR #95, opened 2025-01-29
Add pubkey checks

roshanrags:
The attestation should be bound to the Scallop key. Otherwise, anybody can generate a real attestation and use their own key to set up Scallop.


vg-27 (approved):
LG

PR #96, opened 2025-01-30
Tree memory issue resolved

rajatlko13:
1. assembly code updated - `_selectedPathTree`(dynamic array) memory pointer updated from 0x60 to fmr(0x80)
2. create job allowed without secret

### DIFF
--- a/contracts/contracts/contracts/secret-storage/Executors.sol
+++ b/contracts/contracts/contracts/secret-storage/Executors.sol
@@ -120,7 +120,7 @@ contract Executors is
     error ExecutorsNotTeeManager();
     /// @notice Thrown when the provided execution environment is not supported globally.
     error ExecutorsUnsupportedEnv();
-    error ExecutorsUnavailableStores();
+    error ExecutorsUnavailableResources();
 
     modifier isValidEnv(uint8 _env) {
         _isValidEnv(_env);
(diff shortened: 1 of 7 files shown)


contracts/contracts/contracts/secret-storage/Executors.sol
```diff
@@ -304,28 +304,45 @@ contract Executors is
...
+            // if reqd executors aren't available, then revert
+            if (selectedExecutors.length < noOfExecutorsToSelect)
+                revert ExecutorsUnavailableResources();
+
+            for (uint256 i = 0; i < noOfExecutorsToSelect; i++)
+                selectedNodes[storesCount + i] = selectedExecutors[i];
+        }
 
-        return topNStores;
+        return noOfExecutorsToSelect > 0 ? selectedNodes : topNStores;
```
vg-27, line 328: we can think of optimizing this function, looks like lot of memory copy is happening. One optimization could be, returning the 2 arrays, one for executors_with_secret and another for executor_without_secret. This will save you from joining these 2 arrays in memory.  

another opt can be: in _getTopNstore, you have reordered the selected store array in place, now you can just return the index until which stores are selected.

contracts/contracts/contracts/secret-storage/Executors.sol
```diff
@@ -358,28 +375,28 @@ contract Executors is
...
+        for (uint256 index = 0; index < len; index++) {
+            _deleteIfPresent(_env, _enclaveAddresses[index]);
+        }
+    }
+
+    function _addTreeNodes(
+        uint8 _env,
+        address[] memory _enclaveAddresses
+    ) internal {
+        uint256[] memory stakeAmounts = TEE_MANAGER.getTeeNodesStake(_enclaveAddresses);
```
vg-27, line 392: would it be more costly if we call get stake for each enclave independently, instead of creating a memory array?

contracts/contracts/test/secret-storage/Executors.ts
```diff
@@ -868,25 +878,17 @@ describe("Executors - Select/Release/Slash", function () {
...
-
-    //     expect((await executors.executors(addrs[15])).activeJobs).to.be.eq(1);
-    //     expect((await executors.executors(addrs[16])).activeJobs).to.be.eq(1);
-    // });
+    it("can select executors along with the already selected stores", async function () {
+        await teeManager.drainTeeNode(addrs[16]);
+
+        let env = 1,
+            noOfNodesToSelect = 2;
+        await expect(executors.selectExecutionNodes(env, [addrs[15]], noOfNodesToSelect))
```
vg-27, line 886: can we validate the return values and tree state is correct after this function is called, because tree changes were made in the function.

contracts/contracts/test/secret-storage/job-allocation/Jobs.ts
```diff
@@ -873,43 +873,31 @@ describe("Jobs - Create", function () {
...
+        // Since it is a first job.
+        let jobId = 0;
+        let job = await jobs.jobs(jobId);
+
+        expect(job.jobOwner).to.eq(jobOwner);
+        expect(job.deadline).to.eq(deadline);
+        expect(job.execStartTime).to.eq((await tx.getBlock())?.timestamp);
+        expect(job.env).to.eq(env);
+
+        let selectedExecutors = await jobs.getSelectedExecutors(jobId);
```
vg-27, line 896: validate selectedExecutor length and elements are unique within it.

vg-27 (changes requested):

contracts/contracts/test/secret-storage/Executors.ts
rajatlko13: we cannot validate the return values as it's not a read function, but a txn. 
Also, what tree state changes are you referring to here?

contracts/contracts/contracts/secret-storage/Executors.sol
rajatlko13: yeah it would be more costly execute multiple external  contract calls

rajatlko13: yeah, multiple external contract calls would be more costly

contracts/contracts/test/secret-storage/Executors.ts
vg-27: > we cannot validate the return values as it's not a read function, but a txn.

Find a way to test that selected nodes are as expected, without that UT will be incomplete 

> what tree state changes are you referring to here?

we have delete and add tree node function called inside the selectExecutionNodes, talking about that.

rajatlko13: added ExecutorsUser contract to test the return values from the txn
added validations for tree state changes

contracts/contracts/test/secret-storage/Executors.ts
```diff
@@ -923,8 +936,9 @@ describe("Executors - Select/Release/Slash", function () {
     it("can slash executor", async function () {
         await executors.selectExecutionNodes(1, [addrs[15]], 1);
 
-        await expect(executors.slashExecutor(addrs[15]))
-            .to.be.not.reverted;
+        await expect(executorsUser.slashExecutor(addrs[15]))
+            .to.emit(executorsUser, "ExecutorsUserNodeSlashed")
+            .withArgs(0);
```
vg-27, line 941: can you verify that `TeeManagerMockExecutorSlashed` is emitted? Also you may return some non-zero slashAmount value in the mock.

contracts/contracts/contracts/serverless-v2/tree/TreeMapUpgradeable.sol
```diff
@@ -368,14 +368,10 @@ contract TreeMapUpgradeable is Initializable {
         if (_N > _nodeCount) _N = _nodeCount;
         if (_N == 0) return new address[](0);
 
-        // WARNING - don't declare any memory variables before this point
```
vg-27, line 371: can you put back all comments ?

rajatlko13: done

contracts/contracts/test/secret-storage/Executors.ts
rajatlko13: done

vg-27 (commented):
LG.

vg-27 (approved):
Approving, just update the PR description to mention about the new assembly fix.

PR #97, opened 2025-01-30
pass pcr json path to verify command

Naman-Monga:
using clap[conflicts & requires] in sub args - for clean arg set rules.

### DIFF
--- a/cli/oyster-cvm/src/args/pcr.rs
+++ b/cli/oyster-cvm/src/args/pcr.rs
@@ -56,9 +56,7 @@ impl PcrArgs {
                 )))
             }
             None => {
-                // If any PCR is provided, all must be provided
-                let has_any = self.pcr0.is_some() || self.pcr1.is_some() || self.pcr2.is_some();
-                if !has_any {
+                if self.pcr0.is_none() {
                     return Ok(None);
                 }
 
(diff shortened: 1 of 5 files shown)


cli/oyster-cvm/src/args/pcr.rs
```diff
@@ -0,0 +1,73 @@
...
+                    lower_keys_map
+                        .get("pcr2")
+                        .and_then(|v| v.as_str())
+                        .ok_or(anyhow!("Missing PCR2"))?
+                        .into(),
+                )))
+            }
+            None => {
+                // If any PCR is provided, all must be provided
+                let has_any = self.pcr0.is_some() || self.pcr1.is_some() || self.pcr2.is_some();
```
roshanrags, line 60: is this not handled by `requires_all`?

Naman-Monga: it's handled, put there to check if no value was passed, but yeah, simplified so that there's no confusion.

roshanrags: maybe just add a comment there explaining why one check is enough

Naman-Monga: added

roshanrags (approved):

PR #99, opened 2025-02-02
Structured init params

roshanrags:
Main changes
- Encoder and decoder for init params added, to make them structured and more flexible
- Encrypted init params are supported
- `extra-init-params` is more or less useless
- `vet` exits with an error if status code is not 200
- The Docker SDK uses a staged setup.sh where programs start at different points in the setup.sh script
- Docker SDK includes the KMS derive server
- X25519 keys are the identity instead of Ed25519, latter doesn't really seem to be needed anywhere


roshanrags:
@prateekreddy try using this for Eliza...

initialization/init-params-manager/src/decoder.rs
```diff
@@ -0,0 +1,199 @@
...
+
+    // parse init params
+    let init_params_str = read_to_string(args.init_params_path)
+        .context("failed to read init params, should never happen")?;
+    if init_params_str == "" {
+        // ignore empty params
+        info!("empty init params");
+        return Ok(());
+    }
+    let init_params = serde_json::from_str::<InitParams>(&init_params_str)
```
vg-27, line 57: base64 decoding? is operator expected to do that during init-server setup?

initialization/init-params-manager/src/decoder.rs
```diff
@@ -0,0 +1,199 @@
...
+            if PathBuf::from(&init_param.path)
+                .components()
+                .any(|x| !matches!(x, Component::Normal(_)))
+            {
+                bail!("invalid path")
+            }
+
+            let path = PathBuf::from("/init-params/".to_owned() + &init_param.path);
+            // ensure all files are contained within this directory
+            if !path.starts_with(Path::new("/init-params/")) {
```
vg-27: isn't this redundant if /init-params is getting prepended at L83.

initialization/init-params-manager/src/decoder.rs
roshanrags: yes, the control plane does the first level of base64 encoding so the enclave sees the json directly

initialization/init-params-manager/src/encoder.rs
```diff
@@ -0,0 +1,226 @@
...
+                should_decrypt: should_encrypt,
+            };
+
+            Ok(init_param)
+        })
+        .collect::<Result<Vec<_>>>()
+        .context("failed to build init params")?;
+
+    // create final init params
+    let init_params = InitParams {
```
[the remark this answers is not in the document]
roshanrags: done

vg-27 (commented):
Rename in encoder.rs too, similar to decoder.rs

roshanrags:
done

vg-27 (approved):
LG

PR #100, opened 2025-02-03
Add deposit command to cli

Naman-Monga:


roshanrags:
Add contexts to the errors

PR #108, opened 2025-02-06
Introduce image IDs to fold in PCRs and user data details

roshanrags:
Simplifies everything and also makes it flexible towards change in the future.

### DIFF
--- a/attestation/verifier-risczero/host/src/main.rs
+++ b/attestation/verifier-risczero/host/src/main.rs
@@ -1,6 +1,6 @@
 use clap::Parser;
 use methods::{GUEST_ELF, GUEST_ID};
-use risc0_zkvm::{default_prover, ExecutorEnv, ProverOpts};
+use risc0_zkvm::{default_prover, is_dev_mode, ExecutorEnv, ProverOpts};
 
 #[derive(Parser, Debug)]
 #[command(author, version, about, long_about = None)]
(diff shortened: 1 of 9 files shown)


roshanrags:
Actually, not clear how this works with e.g. solana, unless we use keccak there as well. And if we have to choose something neutral/common across chains, might as well just go with sha256. Wdyt @vg-27 ?

roshanrags:
Switched it to sha256, the risczero verifier has an accelerated version as well

attestation/verifier-risczero/host/src/main.rs
```diff
@@ -49,7 +49,13 @@ fn main() {
 
     println!(
         "Seal: {}",
-        hex::encode(&receipt.inner.groth16().unwrap().seal)
+        hex::encode(
+            receipt
+                .inner
+                .groth16()
+                .map(|x| x.seal.as_slice())
+                .unwrap_or(b"probably dev mode?")
```
vg-27, line 57: message can be improved, to mention the action for caller. here dev mode refers to which entity? Also can we check programmatically if it's in dev mode or not here. And print accordingly.

roshanrags: it won't be visible in plaintext anyway since it is hex encoded, though ig we can just detect dev mode and not even do this

vg-27 (approved):
LG

PR #111, opened 2025-02-18
Add log stream post deployment in debug mode

Naman-Monga:
streams from start after successful deploy. optional arg: `--no-stream` to not stream logs.

### DIFF
--- a/cli/oyster-cvm/src/main.rs
+++ b/cli/oyster-cvm/src/main.rs
@@ -101,6 +101,10 @@ enum Commands {
         #[arg(long)]
         debug: bool,
 
+        /// Disable automatic log streaming in debug mode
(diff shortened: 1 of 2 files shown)


cli/oyster-cvm/src/main.rs
```diff
@@ -101,6 +101,10 @@ enum Commands {
         #[arg(long)]
         debug: bool,
 
+        /// Disable automatic log streaming in debug mode
+        #[arg(long, requires = "debug")]
```
prateekreddy, line 105: default for no_stream can be `false`?

Naman-Monga: it's already false by default.

prateekreddy (approved):

PR #117, opened 2025-03-06
Update KMS root and derive server to use contract validation

vg-27:
- In root-server, replace validation by enclave public key with hash of PCRs and user data
- Update derive server to fetch seed from KMS root-server-contract, optionally.
- Update oyster-cvm cli to fill contract address and KMS contract root server details (endpoint, PCRs, user data)


roshanrags:
Do we want to include it in the blue image itself or make a new base image

vg-27:
> Do we want to include it in the blue image itself or make a new base image

We are starting contract based derive-server conditionally, only if certain input parameters are present. This will save us managing a separate image.

cli/oyster-cvm/src/args/init_params.rs
```diff
@@ -60,6 +77,21 @@ impl InitParamsArgs {
             .unwrap_or_default();
         init_params.append(&mut self.init_params.unwrap_or_default());
 
+        if let Some(address) = self.contract_address {
+            if let Err(_) = Address::from_hex(&address) {
+                bail!("invalid contract address");
+            }
+            init_params.push(format!("contract-address:1:0:utf8:{}", address));
```
roshanrags, line 84: might want to put this right after the compose file instead

cli/oyster-cvm/src/args/init_params.rs
```diff
@@ -266,3 +298,23 @@ fn fetch_encryption_key(
         .try_into()
         .context("failed to parse reponse")
 }
+
+// TODO fill store KMS details
+lazy_static! {
+    static ref KMS_ROOT_SERVERS: HashMap<u64, &'static str> = {
+        let mut root_servers = HashMap::new();
+        root_servers.insert(
+            31337,
```
roshanrags, line 307: hardhat chain id?

vg-27: put there temporarily, will fill it with actual values after the KMS deployment.

roshanrags (approved):

PR #119, opened 2025-03-10
Update serverless executor for code contract changes

Ayush-Yadav:


serverless/executor/src/workerd.rs
```diff
@@ -85,35 +87,42 @@ pub async fn create_code_file(
     }
 
     // Get the calldata of the transaction as string
-    let calldata = match tx_data["input"].take() {
-        Value::String(calldata) => Ok(calldata),
+    let input = match tx_data["input"].take() {
```
vg-27, line 90: update comment the input is hexadecimal string starting with '0x'

vg-27 (commented):
https://docs.rs/ethers/latest/ethers/core/abi/struct.Function.html#method.decode_input can use this helper, would have saved from manual decoding.

serverless/executor/src/workerd.rs
```diff
@@ -86,31 +84,49 @@ pub async fn create_code_file(
...
-    }
-
     // Ensure the function selector is correct
     if !input.starts_with(SAVE_CODE_FUNCTION_SELECTOR) {
         return Err(ServerlessError::InvalidTxCalldata);
     }
+    let input_bytes = hex::decode(&input[2..])?;
+
+    // Create a Function object corresponding to function saveCodeInCallData(string calldata inputData, bytes calldata metadata)
+    let save_code_function = Function {
```
vg-27, line 100: can you use abigen macro to get this function object ?  That would dev friendly, since if there is any changes in contract later then macro has to be updated only.

serverless/executor/src/workerd.rs
```diff
@@ -86,31 +84,49 @@ pub async fn create_code_file(
...
+        outputs: vec![],
+        constant: Some(false),
+        state_mutability: StateMutability::NonPayable,
+    };
 
-    // Get the input bytes starting from the parameter length byte
-    let input_bytes: &[u8] = &hex::decode(&input[138..])?;
-    let code_data_length = U256::from_big_endian(&input_bytes[0..32]);
+    // Now decode the data
+    let Ok(tokens) = save_code_function.decode_input(&input_bytes[4..]) else {
```
vg-27, line 120: why are you stripping the function selector? Also you can skip the function selector check before, this method should do that.

vg-27 (changes requested):

serverless/executor/src/workerd.rs
Ayush-Yadav: This method doesn't take into account the function selector bytes. Tried with the entire input bytes but then the decoding was failing.

vg-27: https://docs.rs/ethers/latest/ethers/contract/fn.decode_function_data.html this looks more elegant then. Check this.

Ayush-Yadav: It does the same thing.

serverless/executor/src/workerd.rs
Ayush-Yadav: Checked, Can't get the function object from the abigen macro generated object.

serverless/executor/src/workerd.rs
vg-27: Agree, but abstraction should be used if it's available instead for doing things ourselves.

vg-27 (dismissed):

serverless/executor/src/workerd.rs
vg-27: Ack.

vg-27 (approved):

Issue #66, opened 2025-01-10
Fix control plane provider rate limiting on startup


roshanrags:

roshanrags:
More or less fixed by #128

PR #128, opened 2025-03-20
Maintain a termination list to avoid iterating through all jobs over and over

roshanrags:
Not perfect but should fix most of #66 

Also removed extra params entirely since it is no longer used.


vg-27 (approved):
LG

PR #138, opened 2025-04-01
sign public endpoint

aniketbaliyan:


roshanrags:
put the signature in a X-KMS-Signature header so the response doesn't have to change

kms/root-server/src/main.rs
```diff
@@ -230,6 +246,10 @@ async fn run_public_server(app_state: AppState, listen_addr: String) -> Result<(
             "/derive/x25519/public",
             get(derive_public::derive_x25519_public),
         )
+        .route(
+            "/get/public",
+            get(derive_public::get_public),
+        )
```
roshanrags, line 252: don't need this, the creator outputs the public key anyway

kms/root-server/src/main.rs
```diff
@@ -91,6 +97,7 @@ async fn main() -> Result<()> {
     // sleep to allow attestation verification
     // taco decryption will only work after the signer is
     // verified on chain
+
```
roshanrags, line 100: remove

roshanrags (commented):
Use secp256k1 keys derived using `to_secp256k1_public(derive_path_seed(
        *message,
        b"oyster.kms.secp256k1",
    ))

roshanrags:
> This is the verification code against which it was tested from the client side

Where is this from?

roshanrags:
Also, this has to be implemented for root-server-contract as well

aniketbaliyan:
The verification code is on my local, just wanted to share here

roshanrags:
superseded by #143

PR #139, opened 2025-04-01
KMS root server hardening

roshanrags:
Added timeouts and other checks where appropriate, if there is anything else that should be added, now is the time. Updated dependencies as well.

Technically, I think we can set request body sizes to zero since everything is GET and we shouldn't be having bodies, I set it to a low value for now just in case it breaks something.

### DIFF
--- a/sdks/rs/src/axum.rs
+++ b/sdks/rs/src/axum.rs
@@ -1,10 +1,12 @@
 use std::net::SocketAddr;
+use std::time::Duration;
 
 use axum::{
     extract::connect_info::Connected,
     serve::{IncomingStream, Listener},
 };
 use tokio::net::{TcpListener, TcpStream};
+use tokio::time::sleep;
 
 use crate::scallop::{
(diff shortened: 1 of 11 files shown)


sdks/rs/src/axum.rs
```diff
@@ -36,15 +40,20 @@ where
...
+            stream = new_server_async_Noise_IX_25519_ChaChaPoly_BLAKE2b(
+                stream,
+                &self.secret,
+                Some(self.auth_store.clone()),
+                Some(self.auther.clone()),
+            ) => {
+                Ok((stream?, addr))
+            }
+            _ = sleep(Duration::from_secs(5)) => {
+                Err(AxumError::TimeoutError)
```
vg-27, line 54: Does this timeout overlap with the timeout layer applied on axum ? If I understand correctly, we are setting connection timeout (includes attestation verification) to 5 sec and the timeout for request handling to 5 secs, which starts after connection is established.

vg-27 (approved):

sdks/rs/src/axum.rs
roshanrags: yes, it should be happening one after the other

PR #140, opened 2025-04-02
Authenticate the KMS root servers with a static derived key

roshanrags:
Replaces the measurements based validation so the KMS root server can be updated and still work with existing enclaves. Security is now delegated to the KmsRoot contract to protect who can derive the static key by accessing the seed, which it does already today. So we get this flexibility effectively for free without changing the security posture.

### DIFF
--- a/kms/derive-server/Cargo.toml
+++ b/kms/derive-server/Cargo.toml
@@ -5,19 +5,19 @@ edition = "2021"
 license = "Apache-2.0"
 
 [dependencies]
-anyhow = "1.0.95"
-axum = "0.8.1"
-clap = { version = "4.5.27", features = ["derive"] }
+anyhow = "1.0.97"
+axum = "0.8.3"
+clap = { version = "4.5.35", features = ["derive"] }
 hex = "0.4.3"
-http-body-util = "0.1.2"
-hyper = { version = "1.5.2", features = ["full"] }
-hyper-util = { version = "0.1.10", features = ["tokio"] }
+http-body-util = "0.1.3"
+hyper = { version = "1.6.0", features = ["full"] }
+hyper-util = { version = "0.1.11", features = ["tokio"] }
 kms-derive-utils = { version = "0.1.0", path = "../derive-utils" }
-oyster-sdk = "0.15.0"
-reqwest = "0.12.12"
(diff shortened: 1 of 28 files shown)


kms/derive-server/Cargo.toml
```diff
@@ -5,19 +5,19 @@ edition = "2021"
...
-hyper = { version = "1.5.2", features = ["full"] }
-hyper-util = { version = "0.1.10", features = ["tokio"] }
+http-body-util = "0.1.3"
+hyper = { version = "1.6.0", features = ["full"] }
+hyper-util = { version = "0.1.11", features = ["tokio"] }
 kms-derive-utils = { version = "0.1.0", path = "../derive-utils" }
-oyster-sdk = "0.15.0"
-reqwest = "0.12.12"
-serde = { version = "1.0.217", features = ["derive"] }
+oyster-sdk = "0.15.3"
```
vg-27, line 16: why not 0.16.0 ?

vg-27 (approved):
After this, in the next PR blue images needs to be updated and oyster-cvm cli.

Also, for KMS contract root server, we can say now that for different chains, seed can remain same hence the kms_pubkey will be same.

kms/derive-server/Cargo.toml
roshanrags: updated, was left over from the previous version

kms/derive-server/src/main.rs
```diff
@@ -25,6 +25,14 @@ struct Args {
     )]
     kms_endpoint: Option<String>,
 
+    /// KMS pubkey, hex encoded
+    #[arg(
+        long,
+        conflicts_with = "root_server_config",
+        required_unless_present = "root_server_config"
+    )]
+    kms_pubkey: Option<String>,
```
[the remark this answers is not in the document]
roshanrags: done

roshanrags:
For the contract kms servers, it is kinda intentional that different chains have different seeds. For one, it namespaces the contracts though this can be made part of the derivation path. Main reason is they depend on RPCs for security so it effectively limits the blast radius in case a specific chain's RPC (or the chain itself) gets compromised.

vg-27 (approved):
LG

PR #143, opened 2025-04-04
Add signatures to public endpoints

roshanrags:


vg-27 (approved):
LG. To verify the signing key, client has to rely on creator-verifier. No attestations required from KMS root-servers. Right?

roshanrags:
> LG. To verify the signing key, client has to rely on creator-verifier. No attestations required from KMS root-servers. Right?

Verifiers (like oyster-cvm) are expected to just hardcode the verification key/address which would basically correspond to a given expected seed.

PR #154, opened 2025-04-11
Update cli to use the new blue image

roshanrags:


vg-27 (dismissed):
- can log which pcr preset is being used for user visibility
- In verification, user has to use --pcr-preset explicitly, because of v1->v2 shift of default preset
- need rebase from master

PR #160, opened 2025-04-21
Create local development image for Oyster enclaves

Ayush-Yadav:
Design Doc: https://docs.google.com/document/d/1KNQHdXcKl2MqS5dK0M6SdBJNE2IQLjcVHUjOkFIrm8Y/edit?usp=sharing


local/dev-image/docker-compose.yml
roshanrags, line 1: what is this for?

local/dev-image/default.nix
```diff
@@ -0,0 +1,85 @@
...
+  derive-server-mock,
+  compose ? ./. + builtins.getEnv "COMPOSE",
+  dockerImages ? let
+    envValue = builtins.getEnv "DOCKER_IMAGES";
+  in if envValue == "" then [ ] else builtins.fromJSON (envValue),
+}: let
+  system = systemConfig.system;
+  pkgs = nixpkgs.legacyPackages."${system}";
+
+  baseImage = if system == "x86_64-linux" then
```
roshanrags, line 16: is a base image needed? what for?

local/dev-image/default.nix
```diff
@@ -0,0 +1,85 @@
+{
+  nixpkgs,
+  systemConfig,
+  supervisord,
+  keygen, 
+  attestation-server-mock, 
+  derive-server-mock,
+  compose ? ./. + builtins.getEnv "COMPOSE",
```
roshanrags, line 8: this is the user's compose file right? shouldn't it be given at runtime?

local/dev-image/default.nix
```diff
@@ -0,0 +1,85 @@
...
+  dockerImages ? let
```
roshanrags, line 9: same for this

local/dev-image/docker-compose.yml
Ayush-Yadav: Just a sample compose file the image can be built with.

local/dev-image/default.nix
Ayush-Yadav: For running the binaries, supervisord, etc. ? Not much needed though that's why have used the minimal alpine image

local/dev-image/default.nix
Ayush-Yadav: Yes In the simulation the CLI command mounts the user's compose file inside the base image container and the docker images as well. This is so that if someone wants to just test the dev image with a docker compose then they can do so.

roshanrags: uhh not sure I understand, why does this even exist at build time in the Nix file if it is supposed to be used at runtime

roshanrags: can't I use whatever docker compose I want at runtime to test

local/dev-image/default.nix
roshanrags: you sure it's needed? you can add additional dependencies in the `buildEnv` block btw

local/dev-image/default.nix
Ayush-Yadav: Yes you can. The container will need to be run with `-v docker-compose.yml:/app/docker-compose.yml` and similarly for docker images.
It is added at build time basically for testing purpose like if someone is updating the base image code then they can build it with their sample docker compose and just run the container to test it.

local/dev-image/default.nix
Ayush-Yadav: Yes can replace that with dependencies like busybox. I thought there won't be much size difference but sure will add the required dependencies instead of the base image.

local/dev-image/default.nix
Ayush-Yadav: Should I remove it?

roshanrags: yeah remove

local/dev-image/docker-compose.yml
roshanrags, line 1: remove

local/dev-image/default.nix
```diff
@@ -0,0 +1,54 @@
...
+    cp ${deriveServerMock} $out/app/kms-derive-server
+    cp ${setup} $out/app/setup.sh
+    chmod +x $out/app/*
+    cp ${supervisorConf} $out/etc/supervisord.conf
+  '';  
+
+in {
+  default = pkgs.dockerTools.buildImage {
+    name = "marlinorg/local-dev-image";
+    tag = "latest";
```
roshanrags, line 41: remove

roshanrags (approved):

PR #168, opened 2025-04-24
Support verifying hex attestations

roshanrags:


cli/oyster-cvm/src/commands/verify.rs
```diff
@@ -11,11 +11,21 @@ use crate::configs::global::DEFAULT_ATTESTATION_PORT;
...
+/// For verifying a running enclave (among other flags):
+///   --enclave-ip <ENCLAVE_IP>
+///
+/// For verifying an existing attestation (among other flags):
+///   --attestation-hex <ATTESTATION_HEX>
 #[derive(Args)]
 pub struct VerifyArgs {
+    /// Hex encoded attestation
+    #[arg(short = 'x', long, conflicts_with = "enclave_ip")]
+    attestation_hex: Option<String>,
```
vg-27, line 24: attestation bytes are usually very long, and it won't look good on cli. Instead, can read it from a file.

cli/oyster-cvm/src/commands/verify.rs
```diff
@@ -11,11 +11,21 @@ use crate::configs::global::DEFAULT_ATTESTATION_PORT;
...
+
     /// Enclave IP
-    #[arg(short = 'e', long, required = true)]
-    enclave_ip: String,
+    #[arg(short = 'e', long, conflicts_with = "attestation_hex")]
```
vg-27, line 27: can use [required_unless_present](https://docs.rs/clap/latest/clap/struct.Arg.html#method.required_unless_present) , so that at least one of them is provided.

vg-27 (approved):

cli/oyster-cvm/src/commands/verify.rs
roshanrags: oh files are kinda inconvenient to me lol :laughing: added this mainly so users can verify the kms seed attestation by copy pasting from docs

roshanrags:
@vg-27 does this also make the attestation verifier binary in `attestation/verifier` redundant?

cli/oyster-cvm/src/commands/verify.rs
roshanrags: added file option as well

vg-27 (dismissed):
LG

vg-27:
> @vg-27 does this also make the attestation verifier binary in `attestation/verifier` redundant?

Yeah, true.

vg-27:
Need readme update for this too.

PR #176, opened 2025-04-29
Utilities for extending and locking custom PCRs

roshanrags:


initialization/pcr-utils/src/locker.rs
```diff
@@ -0,0 +1,36 @@
...
+    api::{Request, Response},
+    driver::{nsm_exit, nsm_init, nsm_process_request},
+};
+use clap::Parser;
+
+/// Lock PCRs
+#[derive(Parser)]
+#[command(author, version, about, long_about = None)]
+struct Args {
+    /// PCR index, should be within [16, 31] inclusive
```
vg-27, line 12: shouldn't we enforce the range?

vg-27 (approved):

initialization/pcr-utils/src/locker.rs
roshanrags: tried it in clap, doesn't seem to work properly, the NSM API will do it anyway

PR #180, opened 2025-05-02
Fetch response + code refactor

sagarparker:


cli/oyster-serverless/README.md
```diff
@@ -150,4 +150,18 @@ oyster-serverless subscription create
...
+
+
+```bash
+oyster-serverless subscription fetch-response
+```
+
+**Arguments:**
+
+- `subscription-transaction-hash`: Transaction hash returned by the create subscription command.
+- `stream` : Stream the response
```
vg-27, line 167:
```suggestion
- `stream` : If set, listen for job responses until subscription is terminated. Else, get the responses received till the current time. 
```

cli/oyster-serverless/src/commands/subscription/fetch.rs
```diff
@@ -0,0 +1,149 @@
...
+            }
+
+            // Sleep for a shorter duration since we're tracking blocks explicitly
+            tokio::time::sleep(std::time::Duration::from_secs(5)).await;
+        }
+    } else {
+        // Non-stream mode: fetch all existing responses at once
+        info!("Fetching all existing responses...");
+
+        let latest_block = provider.get_block_number().await?;
```
vg-27, line 137: for to_block field in event filter, it accepts 'BlockNumberOrTag:Latest' , which means you don't have to explicitly get latest block number and fill it, rpc will itself provide logs until latest block.
can save on one API.

vg-27 (dismissed):

cli/oyster-serverless/src/commands/subscription/fetch.rs
```diff
@@ -115,8 +113,7 @@ pub async fn fetch_response(args: FetchResponseArgs) -> Result<()> {
             fetch_and_process_subscription_events(
```
vg-27, line 113: for consistency, this function should return the latest_block, instead of using it from a independent get_block_number call. Because they can be different.

sagarparker: lets just stick to the older version

vg-27 (approved):

PR #183, opened 2025-05-05
Terminate and refund subscription commands

sagarparker:


cli/oyster-serverless/src/commands/subscription/terminate.rs
```diff
@@ -0,0 +1,112 @@
...
+        .context("Failed to get transaction receipt")?;
+
+    info!(
+        "Transaction hash for terminate subscription transaction: {:?}",
+        terminate_receipt.transaction_hash
+    );
+
+    info!(
+        "If you plan to execute the refund-deposits command next, please wait at least {} seconds.",
+        overall_timeout
```
vg-27, line 108: Can you make this 2 * overall_timeout ? 
While going through the contracts, found out that, when terminating, termination happens after overall_timeout. And refund can be carried out after overall_timeout of actual termination. So it's 2 times of overall_timeout.

vg-27 (approved):

PR #204, opened 2025-05-20
 Update for `new` and `deploy` command for the updated mainnet code contract

sagarparker:


cli/oyster-serverless/src/configs/global.rs
```diff
@@ -1,6 +1,8 @@
+// TODO : Change contract address for code contract to mainnet and RPC URL to mainnet
```
vg-27, line 1: can remove the TODO

vg-27 (dismissed):

cli/oyster-serverless/src/configs/global.rs
sagarparker: updated

PR #210, opened 2025-06-12
Migrate executor code from ethers crate to alloy and incorporate transaction-manager for sending transactions

Ayush-Yadav:


serverless/executor/src/execution.rs
```diff
@@ -52,12 +51,7 @@ pub async fn handle_job(
     let _ = workerd::cleanup_code_file(&code_hash, slug, &app_state.workerd_runtime_path).await;
 
     // Initialize the default timeout response and build on that based on the above response
-    let mut job_output = Some(JobOutput {
-        output: Bytes::new(),
-        error_code: 4,
-        total_time: user_deadline.into(),
-        ..Default::default()
-    });
+    let mut job_output = Some((Bytes::new(), 4, user_deadline.into()));
```
vg-27, line 54: isn't using tuple confusing?

serverless/executor/src/model.rs
```diff
@@ -83,75 +96,26 @@ pub struct MutableConfig {
...
 
-#[derive(Serialize)]
+#[derive(Deserialize, Serialize)]
 pub struct RegistrationMessage {
     pub job_capacity: usize,
     pub storage_capacity: usize,
     pub sign_timestamp: u64,
-    pub env: u8,
-    pub owner: H160,
+    pub env: U256,
```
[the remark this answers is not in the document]
Ayush-Yadav: Ah I guess that's by mistake. Had updated the constant for env id to U256 for simplicity in adding to the events filter topic.

serverless/executor/src/events.rs
```diff
@@ -0,0 +1,394 @@
...
+                );
+                continue;
+            }
+        };
+        let jobs_created_stream = pin!(jobs_created_stream.into_stream());
+
+        // Create filter to listen to JobResponded events emitted by the Jobs contract
+        let jobs_responded_filter = Filter::new()
+            .address(app_state.jobs_contract_addr)
+            .event(JobsContract::JobResponded::SIGNATURE)
```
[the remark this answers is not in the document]
Ayush-Yadav: Yeah that'll be better. Will add once the contract is updated.

serverless/executor/src/execution.rs
Ayush-Yadav: Yes it is a bit. Will create a struct then.

serverless/executor/src/events.rs
```diff
@@ -210,26 +211,30 @@ pub async fn handle_event_logs(
...
-                if event.topic0() == Some(&ExecutorsContract::TeeNodeDeregistered::SIGNATURE_HASH) {
-                    println!("Executor deregistered from the common chain!");
+                match event.topic0() {
+                    // Capture the Enclave deregistered event emitted by the 'TeeManager' contract
+                    Some(&TeeManagerContract::TeeNodeDeregistered::SIGNATURE_HASH) => {
+                        println!("Executor deregistered from the common chain!");
                     app_state.enclave_registered.store(false, Ordering::SeqCst);
 
                     println!("Stopped listening to 'Jobs' events!");
                     return;
```
vg-27, line 221: nit: indentation to right, and check further lines too. can use `cargo fmt`

serverless/executor/src/model.rs
```diff
@@ -4,6 +4,7 @@ use std::sync::{Arc, Mutex, RwLock};
 
 use alloy::primitives::{Address, U256};
 use alloy::sol;
+use axum::body::Bytes;
```
vg-27, line 7: using axum here feels out of context, because axum is used in http handlers and servers. which is not used here. Can you use some generic or relavant bytes struct? Use the class based on it's requirement and context. Avoid unnecessary type transformations.

serverless/executor/src/workerd.rs
```diff
@@ -2,7 +2,7 @@ use std::process::Child;
 use std::time::{Duration, Instant};
 
 use alloy::sol_types::SolCall;
-use bytes::Bytes;
+use axum::body::Bytes;
```
vg-27, line 5: check this axum too. Is it added by copilot?

serverless/executor/src/utils.rs
```diff
@@ -101,3 +101,9 @@ pub async fn get_latest_block_number(http_rpc_url: &String) -> Result<u64> {
         .await
         .context("Failed to get a response from the rpc server")
 }
+
+pub fn get_byte_slice(num: u8) -> [u8; 32] {
```
vg-27, line 105: doesn't alloy provide any good abstraction to convert u8, u32, u64 to U256?

serverless/executor/src/events.rs
```diff
@@ -0,0 +1,394 @@
...
+                    .unwrap()
+                    .insert(job_id);
+
+                // Check if the executor has been selected for the job execution
+                let is_node_selected = event_decoded.selectedExecutors.clone()
+                    .into_iter()
+                    .any(|addr| addr == app_state.enclave_address);
+
+                let app_state_clone = app_state.clone();
+                let user_deadline = event_decoded.deadline.clone().as_limbs()[0];
```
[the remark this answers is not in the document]
vg-27: so chunk of 64 bit unsigned int is called limb

serverless/executor/src/model.rs
Ayush-Yadav: Was using `bytes` from the `bytes` crate but felt like it was an overkill. And we'll have to do type transformations when converting bytes from endpoint calls to job output transaction creation which uses alloy bytes. Besides we're just representing the `bytes` data here which is an array of bytes and not doing any operations that might be attached to different class objects.

serverless/executor/src/workerd.rs
Ayush-Yadav: No I had added this myself, also when using `bytes::bytes` as well there were conversions using `.into()` because different parts of code return or ingest different bytes type like alloy, axum, reqwest.

serverless/executor/src/events.rs
```diff
@@ -0,0 +1,394 @@
...
+
+// Start task to handle the execution timeout scenario for a job request
+async fn handle_timeout(
+    job_id: U256,
+    timeout: u64,
+    app_state: State<AppState>,
+    tx_sender: Sender<JobsTransaction>,
+) {
+    sleep(Duration::from_millis(
+        timeout + app_state.execution_buffer_time * 1000 + 1000,
```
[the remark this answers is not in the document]
Ayush-Yadav: Yes this is happening still. Will add the constant.

vg-27 (approved):

PR #212, opened 2025-06-18
debug imageid correction

aniketbaliyan:


cli/oyster-cvm/src/args/init_params.rs
```diff
@@ -164,9 +164,10 @@ impl InitParamsArgs {
 
         // load pcrs
         // use pcrs of the blue base image by default
+        // if debug flag is true, use zero pcrs
         let pcrs = self
             .pcrs
-            .load_required(preset_to_pcr_preset(&preset, &arch))
+            .load_required(preset_to_pcr_preset(if debug { "debug" } else { &preset }, &arch))
```
vg-27, line 170: can you add another arg in preset_to_pcr_preset function and handle that within the function, because  if want to call it again similar way, it will require inline if-else.

vg-27 (dismissed):

vg-27 (approved):

PR #225, opened 2025-08-20
Add retry delays for RPC errors in serverless executor and secret store

Ayush-Yadav:


serverless/executor/src/events.rs
```diff
@@ -66,6 +66,7 @@ pub async fn events_listener(app_state: State<AppState>, starting_block: u64) {
                         app_state.tee_manager_contract_addr,
                         err,
                     );
+                    sleep(Duration::from_millis(100)).await;
```
vg-27, line 69: replace 100 with some constant. Define that constant separately. And let's make it 1 second sleep. Because honestly we are retrying blindly on any error, not looking at failure reason. So, let's slow down the retry process.

vg-27 (approved):

PR #231, opened 2025-09-25
Update CP arch for DB based market events ingestion.

Ayush-Yadav:


operator/control-plane/src/market.rs
```diff
@@ -200,171 +266,158 @@ pub async fn run(
...
+async fn fetch_job_events(
+    pool: &PgPool,
+    last_processed_id: i64,
+) -> Result<Vec<JobEvent>, sqlx::Error> {
+    let events = sqlx::query_as::<_, JobEvent>(
+        r#"
+        SELECT id, job_id, event_name, event_data
+        FROM job_events
+        WHERE id > $1
+        ORDER BY id ASC
```
roshanrags, line 406: should probably add a limit as well, like 1000 maybe

roshanrags:
not convinced that a channel based architecture works well. there's no place to do backpressure no? jobs can get stuck for minutes at a time e.g. spinning up and channels have limited capacity

operator/control-plane/src/market.rs
```diff
@@ -373,83 +426,141 @@ async fn job_manager(
...
+    // The processing loop follows this:
+    // Keep processing events till you hit an unsuccessful processing
+    // If result is Retry or Internal, these are likely RPC issues and bugs
+    // Hence just break out, the parent function handles retrying
+    // If result is Done, the job is naturally "done", schedule termination
+    // If result is Failed, the job ran into a user error, schedule termination
+    // If job is insolvent, schedule termination
+    // Once job is successfully terminated, break out
+    // Insolvency and heartbeats only matter when job is not already scheduled for termination
+    'event: loop {
```
roshanrags, line 451: does job registry state have to be reset?

operator/control-plane/src/market.rs
```diff
@@ -200,171 +266,158 @@ pub async fn run(
 
     // connection level loop
     // start from scratch in case of connection errors
-    // trying to implicitly resume connections or event streams can cause issues
-    // since subscriptions are stateful
+    // trying to implicitly resume connections can cause issues
 
-    let mut job_count = 0;
+    let mut last_processed_id: i64 = -1;
     loop {
```
roshanrags, line 272: job registry state has to be reset?

Ayush-Yadav: why?

operator/control-plane/src/market.rs
```diff
@@ -373,83 +426,141 @@ async fn job_manager(
...
+                        }
+
+                        cur_id = log.id;
+
+                        match parse_event(log.event_name, log.event_data) {
+                            Ok(event) => state.process_event(event, rates, gb_rates, address_whitelist, address_blacklist),
+                            Err(result) => result
+                        }
+                    }
+                    None => JobResult::Internal,
```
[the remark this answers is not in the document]
roshanrags: I mean the `run` function, where all the senders are created.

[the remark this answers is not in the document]
roshanrags: though ig you don't drop the senders at all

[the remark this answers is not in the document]
Ayush-Yadav: They are created inside the DB events processing loop and are stored in the `JobRegistry`. The ones created in the loop are dropped after going out of scope in the next loop run and the ones stored in the registry are dropped or deleted at the end of the `job_manager` task for that job (Line 536).

roshanrags:
I think `openssl` got added in the indexer, either find a way to remove it or make it vendored.

Ayush-Yadav:
> CP already has it no

Ah yes it has, but still openssl is in the dependency tree. Is it because of the vendoring?

roshanrags:
it might be from dev dependencies

Ayush-Yadav:
> Do an `EXPLAIN` on all the queries (both indexer and cp, reads as well as writes) to see if they are using indexes properly. Confirm once done.

Checked all the queries and confirm that they are using the indexes properly.

PR #232, opened 2025-09-29
Create oyster market events indexer framework and implementations for Arb ONE and Sui blockchains.

Ayush-Yadav:
* Generic indexer framework for indexing oyster market 'job' events into a Postgres DB meant for integration with CP.
* Implement the indexer for ARB ONE network using the framework.
* Implement the indexer for Sui network using the framework.

### DIFF
--- a/operator/market-indexer/arb/Cargo.toml
+++ b/operator/market-indexer/arb/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "oyster-arb-market-indexer"
 version = "0.1.0"
-edition = "2021"
+edition = "2024"
 license = "AGPL-3.0-or-later"
 
 [dependencies]
@@ -15,7 +15,6 @@ tokio-retry = "0.3"
 tracing-subscriber = { version = "0.3.19", features = ["env-filter"] }
 
 [profile.release]
-strip = true
 lto = true
 panic = "abort"
 codegen-units = 1
--- a/operator/market-indexer/framework/migrations/20250919140741_init.sql
+++ b/operator/market-indexer/framework/migrations/20250919140741_init.sql
@@ -1,10 +1,23 @@
 -- Up
 
+-- Create ENUM type for event names
+CREATE TYPE event_name AS ENUM (
+'Opened', 
+'Closed', 
+'Deposited', 
+'Settled', 
+'MetadataUpdated', 
+'Withdrew', 
+'ReviseRateInitiated',
+'ReviseRateCancelled',
+'ReviseRateFinalized'
+);
+
 -- Job events table
 CREATE TABLE job_events (
     id BIGSERIAL PRIMARY KEY,
     job_id VARCHAR(66) NOT NULL,
-    event_name VARCHAR(255) NOT NULL,
+    event_name event_name NOT NULL,
     event_data JSONB NOT NULL,
     indexer_process_time TIMESTAMPTZ DEFAULT now()
 );
@@ -29,3 +42,4 @@ CREATE INDEX idx_job_events_event_name_job_id ON job_events (event_name, job_id)
 -- DROP INDEX idx_job_events_event_name_job_id;
 -- DROP TABLE indexer_state;
 -- DROP TABLE job_events;
+-- DROP TYPE event_name;
--- a/operator/market-indexer/framework/src/chain.rs
+++ b/operator/market-indexer/framework/src/chain.rs
@@ -4,7 +4,7 @@ use std::future::Future;
 use anyhow::{Context, Result};
 
 use crate::events::JobEvent;
-use crate::schema::JobEventRecord;
+use crate::schema::{JobEventName, JobEventRecord};
 
 // Define trait for conversion from raw log to structured JobEvent
 pub trait FromLog: Sized {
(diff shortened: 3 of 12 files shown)


roshanrags:
Licenses all need 2025

README.md
```diff
@@ -40,6 +40,7 @@ The repository is organized into projects:
 - [networking/iperf3-enclave](./networking/iperf3-enclave/): Enclave image that packages iperf3 for benchmarking purposes.
 - [operator/control-plane](./operator/control-plane/): Control plane that manages deployments on behalf of Oyster operators.
 - [operator/quota-monitor-aws](./operator/quota-monitor-aws/): Quota monitor to help Oyster operators manage AWS quotas and resource limits.
+- [operator/market-indexer](./operator/market-indexer/): Oyster market events indexing for multiple blockchains into Postgres Database.
```
roshanrags, line 43: If they are all more or less separate projects, list them all separately here

operator/market-indexer/framework/src/chain.rs
roshanrags, line 1: Document the design choices and why the traits look the way they do. It's not really clear why `fetch_logs_and_group_by_block` and `process_logs_in_block` are separate.

operator/market-indexer/framework/src/events.rs
```diff
@@ -0,0 +1,76 @@
+use alloy_primitives::U256;
```
roshanrags, line 1: if events are common across chains, they shouldn't really be using `alloy`

operator/market-indexer/framework/src/repository.rs
```diff
@@ -0,0 +1,172 @@
...
+        let migrator = Migrator::new(Path::new(MIGRATION_PATH))
+            .await
+            .context("Failed to initialize the migrator")?;
+        migrator
+            .run(&self.pool)
+            .await
+            .context("Failed to apply migrations to the database")
+    }
+
+    pub async fn get_active_jobs(&self) -> Result<HashSet<String>> {
```
roshanrags, line 40: is there a point to this?

roshanrags:
always run `cargo clippy` and fix stuff btw

operator/market-indexer/framework/src/schema.rs
```diff
@@ -0,0 +1,16 @@
+use serde_json::Value;
+use sqlx::prelude::FromRow;
+use sqlx::types::chrono::{DateTime, Utc};
+
+/// A structured representation of the data to be inserted into the `job_events` table
+#[derive(Clone, Debug, FromRow)]
+pub struct JobEventRecord {
+    pub block_id: i64,
+    pub tx_hash: String,
```
roshanrags, line 9: tx index is probably more useful than tx hash for sorting

operator/market-indexer/framework/migrations/20250919140741_init.sql
```diff
@@ -0,0 +1,39 @@
+-- Up
+
+-- Job events table
+CREATE TABLE job_events (
+    id BIGSERIAL PRIMARY KEY,
+    block_id BIGINT NOT NULL, -- Sui: checkpoint_sequence_number; Eth/Arbitrum: block_number; Solana: slot_index
+    tx_hash VARCHAR(100) NOT NULL,
```
roshanrags, line 7: tx index is probably more useful than tx hash for sorting

operator/market-indexer/framework/migrations/20250919140741_init.sql
```diff
@@ -0,0 +1,39 @@
...
+
+-- Useful indexes
+CREATE INDEX idx_job_events_polling ON job_events (indexer_process_time);
+CREATE INDEX idx_job_events_full_fetch ON job_events (block_id, event_seq);
+CREATE INDEX idx_job_events_event_name_job_id ON job_events (event_name, job_id);
+
+-- -- Down
+-- --sqlx DOWN
+
+-- DROP INDEX idx_job_events_polling;
```
roshanrags, line 36: missing an index, and order everything in reverse order of creation

operator/market-indexer/framework/src/events.rs
Ayush-Yadav: Yes there were no better options, I chose `alloy-primitives` because it's the minimal crate that contains this type and also supports some default methods required for our use case.

operator/market-indexer/framework/src/chain.rs
roshanrags: why not just get transformed logs directly

Ayush-Yadav: Block log records are meant to be inserted sequentially for proper tracking and therefore had separated the transform computation sequentially as well so that CP gets to see some events while polling in case there are a lot of logs in a fetched block range.

operator/market-indexer/framework/migrations/20250919140741_init.sql
roshanrags: lots of events will have the same process time no?

operator/market-indexer/framework/src/chain.rs
Ayush-Yadav: > even outside this, seems like `process_logs_in_block` doesn't really have to be different for different chains does it?

Yes it isn't much different except for a few differences in transforming fields like `txn_hash`, `sender`, `block_timestamp` as per chain.

roshanrags: for sui, we plan for `fetch_logs_and_group_by_block` to return _all_ logs including other apps?

operator/market-indexer/framework/migrations/20250919140741_init.sql
Ayush-Yadav: > and even batch inserts will have monotonic ids no?

Yes they will have but I have kept this ordering in the CP queries for making it future proof in case we change the indexer arch later for parallel inserts or something.

operator/market-indexer/framework/src/chain.rs
Ayush-Yadav: My point is we can't parallelize the transformation and we have to insert the records block-wise sequentially in the DB. So we have 2 options:- Transform all the logs for the entire block range first and then sequentially insert them in the DB block-wise or transform them block-wise on the go as they are to be inserted. I don't see the issue with the second approach, if not it's a bit better than the first one because of less lag in the first batch insert.

operator/market-indexer/framework/migrations/20250919140741_init.sql
roshanrags: it's fine don't overengineer

Ayush-Yadav: Okay will use ordering by id then

operator/market-indexer/framework/src/chain.rs
Ayush-Yadav: So should I remove all these fields then?

operator/market-indexer/framework/migrations/20250919140741_init.sql
roshanrags: yeah can remove whatever isn't being used

roshanrags:
use the 2024 edition

operator/market-indexer/framework/migrations/20250919140741_init.sql
```diff
@@ -0,0 +1,31 @@
...
+    id INT PRIMARY KEY,
+    last_processed_block BIGINT NOT NULL,
+    updated_at TIMESTAMPTZ DEFAULT now()
+);
+
+-- Initial values
+INSERT INTO indexer_state (id, last_processed_block) VALUES (1, -1);
+
+-- Useful index (for querying active job_id's based on the JobOpened & JobClosed events)
+CREATE INDEX idx_job_events_event_name_job_id ON job_events (event_name, job_id);
```
roshanrags, line 23: the cp would also probably use some indexes right? if so, are we adding them here or in the cp PR

operator/market-indexer/framework/src/chain.rs
```diff
@@ -0,0 +1,165 @@
...
+
+    /// Fetch raw logs for the oyster market on the chain between a range and group them by block/checkpoint/slot
+    fn fetch_logs_and_group_by_block(
+        &self,
+        start_block: u64,
+        end_block: u64,
+    ) -> impl Future<Output = Result<BTreeMap<u64, Vec<Self::RawLog>>>> + Send;
+
+    // Transform raw logs from a block into suitable DB records
+    fn transform_block_logs_into_records(
```
roshanrags, line 29: doesn't have to be in the trait?

operator/market-indexer/framework/migrations/20250919140741_init.sql
```diff
@@ -0,0 +1,31 @@
+-- Up
+
+-- Job events table
+CREATE TABLE job_events (
+    id BIGSERIAL PRIMARY KEY,
+    job_id VARCHAR(66) NOT NULL,
+    event_name VARCHAR(255) NOT NULL,
```
roshanrags, line 7: maybe use an enum

operator/market-indexer/framework/migrations/20250919140741_init.sql
roshanrags: anyway, can add it in the cp PR based on the query there

roshanrags: random panics :man_shrugging:

operator/market-indexer/arb/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+tracing-subscriber = { version = "0.3.19", features = ["env-filter"] }
+
+[profile.release]
+strip = true
+lto = true
+panic = "abort"
+codegen-units = 1
+
+[profile.dev]
+panic = "abort"
```
roshanrags, line 24: don't think we should strip or abort on panic, same for others as well

operator/market-indexer/framework/migrations/20250919140741_init.sql
roshanrags: tasks can panic for reasons out of our control, like from some library or allocation failures or whatever

Ayush-Yadav: > random panics 🤷‍♂️

In the original design the job manager task only restarted when the last `JobResult` status was `Retry` which only was the case for randomly ended events stream and nothing else

roshanrags: sure, that swallows panics as well I'm guessing

roshanrags: > > tasks can panic for reasons out of our control, like from some library or allocation failures or whatever
> 
> But since CP is such a crucial service the ideal behavior for any such random panic should be to abort and restart it because if it's a random reason then what is the guarantee it won't happen again and again

because initial sync takes time, or used to at least, idk with the current architecture

Ayush-Yadav: > sure, that swallows panics as well I'm guessing

I don't think so, Can you illustrate through an example?

roshanrags: sure, job task panics for whatever reason, task is killed by tokio, the job task now doesn't exist forever till a cp restart

Ayush-Yadav: > > But it was the case originally as well, also what if for that whatever reason the other tasks also starts getting killed? I think we should design a service in such a way that it covers most of the error and corner cases which we have done because the original CP works fine afaik and even then if the service panics then it's best to abort and investigate the issue
> 
> aborting on panics is an option, but only if initial sync is quick (< 10s)

Well that'll depend on the operator, how many jobs it was assigned and how many was terminated things like that, but looking at the current usage and jobs stats I think it should be doable

roshanrags: ditch for now, we'll discuss in the cp pr

operator/market-indexer/arb/Cargo.toml
Ayush-Yadav: Indexer is a quite crucial service and I have handled most of the error cases that can occur. If there are random panics that are not handled because obviously they are random and we don't know about them then aborting on panic is the ideal choice. What if we miss indexing certain events because of it?

roshanrags: have to benchmark initial sync before that

roshanrags: aborting on panics means you don't get good backtraces

roshanrags: all panics will eventually abort unless caught, for tokio, it's only really relevant when you spawn tasks from the main task, then you have to figure out how to handle panics from the children

roshanrags: ig it's fine if we don't want to think about all this, can abort on panics

roshanrags: ah ok

roshanrags: remove stripping anyway

roshanrags (approved):

PR #264, opened 2026-02-11
Add enclave registration on Sui chain and update list command in CVM CLI

Ayush-Yadav:
* Add `register` command for enclave registration on Sui chain via CLI
* Update `list` command for Sui and BSC chains


cli/oyster-cvm/src/commands/list.rs
```diff
@@ -148,34 +192,43 @@ fn process_job_data(node: &Value, now: f64) -> Option<JobData> {
...
+        debug!("Skipping job {} due to zero balance", id);
         return None;
     }
-
-    let last_settled = node
-        .get("lastSettled")
-        .and_then(|v| v.as_str())
-        .and_then(|s| DateTime::parse_from_rfc3339(&format!("{s}Z")).ok())
-        .map(|dt| dt.timestamp() as f64)?;
+    let balance_usdc = balance_raw as f64 / 1_000_000.0;
```
prateekreddy: USDC has 18 decimals in case of BSC so this might not always hold.

cli/oyster-cvm/src/commands/list.rs
```diff
@@ -189,7 +242,7 @@ fn process_job_data(node: &Value, now: f64) -> Option<JobData> {
 
     let current_balance = balance_usdc - (delta_hours * rate_per_hour);
 
-    if current_balance <= 0.0 {
+    if current_balance <= f64::EPSILON {
```
prateekreddy: Is this to account for precision errors? Is that a problem you see regularly?

cli/oyster-cvm/src/commands/list.rs
Ayush-Yadav: Ah Yes, Sorry I missed this. Should I hardcode the decimals per chain to use here or query it via RPC?

cli/oyster-cvm/src/commands/list.rs
Ayush-Yadav: Yes, I was seeing a case where a value like 0.1212 was passing this check even though the value wasn't really <= 0.0

cli/oyster-cvm/src/commands/list.rs
prateekreddy: hardcoding is fine imo since there things don't change and query every time is additional overhead.

cli/oyster-cvm/src/commands/list.rs
prateekreddy: Does replacing 0.0 with EPSILON fix the issue? If so how?

Ayush-Yadav: Sorry I got confused with the check that is before this line regarding the actual balance. So here epsilon was just used to guard against tiny floating-point residuals that can appear after subtraction even when the logical value should be zero.
Using EPSILON treats those near-zero artifacts as zero, so we don’t incorrectly drop a job due to precision noise.

prateekreddy: wdyt about using the scaled up values everywhere except where we are displaying to user? This will help us avoid floats and precision errors?

Ayush-Yadav: Yes sure we can do that. I had done it for the actual balance check and stuff. Will implement for the whole processing.

cli/oyster-cvm/src/commands/list.rs
```diff
@@ -37,35 +43,65 @@ pub async fn list_jobs(args: ListArgs) -> Result<()> {
...
-        }
-    });
+        }),
+    };
 
-    let response = client
-        .post(INDEXER_URL)
-        .json(&query)
+    let mut request = client.post(indexer_url).json(&query);
+    if let Deployment::Bsc = args.deployment {
```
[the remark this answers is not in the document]
Ayush-Yadav: Leaving as-is because the API key is required and will be hardcoded before building a release.

cli/oyster-cvm/src/configs/sui.rs
```diff
@@ -17,3 +17,14 @@ pub const SUI_CLOCK_INITIAL_VERSION: u64 = 1;
...
+// Indexer URL
+pub const INDEXER_URL: &str = "https://sui.indexer.oyster.marlin.org/graphql";
+
+pub const OYSTER_ENCLAVE_REGISTRY_PACKAGE_NAME: &str = "enclave_registry";
+pub const OYSTER_ENCLAVE_REGISTER_METHOD_NAME: &str = "register_enclave";
+// TODO: Replace with mainnet values before release
+pub const OYSTER_ENCLAVE_REGISTRY_PACKAGE_ID: &str =
+    "0xb5a9486ec59fd4355647f6e9fdce5ed282857d9673b8fe4f9cb777e224a163a2";
+pub const REGISTRY_ID: &str = "0x17b03ad8e91e3aab9b71890b40556af9664849b90c2fdb371df6f85fa457634b";
+pub const REGISTRY_INITIAL_VERSION: u64 = 753664486;
```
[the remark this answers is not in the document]
Ayush-Yadav: CLI is only released for the mainnet, the current values are for testing purpose on the testnet. Values will be replaced with mainnet IDs once the review and testing is successful.

cli/oyster-cvm/src/commands/list.rs
```diff
@@ -37,35 +43,65 @@ pub async fn list_jobs(args: ListArgs) -> Result<()> {
 
     info!("Listing active jobs for wallet address: {}", wallet_address);
 
+    let indexer_url = match args.deployment {
+        Deployment::Arbitrum => arb::INDEXER_URL,
+        Deployment::Bsc => bsc::INDEXER_URL,
+        Deployment::Sui => sui::INDEXER_URL,
+    };
```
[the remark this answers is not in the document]
Ayush-Yadav: No issue.

cli/oyster-cvm/src/commands/list.rs
```diff
@@ -37,35 +43,65 @@ pub async fn list_jobs(args: ListArgs) -> Result<()> {
...
                     }
-                    orderBy: CREATED_DESC
-                ) {
-                    nodes {
+                }
+            "#,
+            "variables": {
+                "owner": wallet_address,
+            }
+        }),
```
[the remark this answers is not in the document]
Ayush-Yadav: No error found.

cli/oyster-cvm/src/configs/bsc.rs
```diff
@@ -7,3 +7,9 @@ pub const BSC_RPC_URL: &str = "https://bsc.drpc.org";
 
 // Default operators
 pub const DEFAULT_OPERATOR_ADDRESS: &str = "0xD379eAb71eC709B20A85e953EbD3d395E032a9Af";
+
+// Indexer URL
+pub const INDEXER_URL: &str =
+    "https://gateway.thegraph.com/api/subgraphs/id/AqbdfugmZpgJGrE1ZRbWZ1hLwbX6z9CeKWx3XrzSh72W";
+// TODO: Enter real key before release
+pub const INDEXER_API_KEY: &str = "";
```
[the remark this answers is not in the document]
Ayush-Yadav: It't not upto the user to configure or provide the API key so it isn't being read from an environment variable or CLI flag.

roshanrags (approved):

